PR #3, opened 2021-06-01
remove duplicated/outdated experimental metrics docs

RileyMichael:
This removes the section utilizing `metrics_*` keys under the `experimental` table, based on the assumption they are no longer relevant now that there is a dedicated `metrics` table.

### DIFF
--- a/reference/metrics.html.md
+++ b/reference/metrics.html.md
@@ -11,7 +11,7 @@ nav: firecracker
 
 ## Sending custom metrics to Prometheus
 
-Apps can expose a metrics endpoint we'll periodically (a few times per minute).
+Apps can expose a metrics endpoint we'll periodically poll (a few times per minute).
 
 ### Configuration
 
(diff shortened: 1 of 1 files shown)


reference/metrics.html.md
```diff
@@ -31,28 +31,6 @@ Instrument your app and expose your metrics on `0.0.0.0`.
...
-
-```toml
-app = "your-app-name"
-
-[experimental]
-metrics_port = 9091 # default for most prometheus clients
-metrics_path = "/metrics" # default for most prometheus clients
-```
-
-### Instrumentation
```
mrkurt, line 50: We actually want to keep this section, it's important!

mrkurt (commented):
This is great, one small fix and we'll merge it.

reference/metrics.html.md
RileyMichael: Are you referring to the instrumentation section? If so, it's already present on lines 28-32.

mrkurt: Oh yes, I misunderstood the diff. All good!

PR #5, opened 2021-06-12
Adding/removing flyctl commands

monroeclinton:
Some commands were not listed on the sidebar so I added them. I removed the `flyctl builds` from documentation. I added a page for `flyctl curl` and `flyctl config env`. I removed non-existent commands from the postgres page


tqbf:
Do we want to keep `flyctl curl`?

mrkurt:
We do not want to keep `fly curl`!

monroeclinton:
Removed curl

jsierles:
Thanks for the hard work!

PR #6, opened 2021-06-12
use correct filename for node command

tobiastornros:
Use correct filename for the terminal command when testing out the express server


michaeldwan:
Thank you!

PR #19, opened 2021-08-06
Move copying of config later to avoid full dependency and asset recom…

Maxim-Filimonov:
…pilation

Following the current example, I noticed that every time I update runtime.exs or any other files in the config directory it leads to full recompilation of all dependencies. On a large project, it's a long step. So I recommend moving it down a bit so that we only use it when we do the actual compilation.

### DIFF
--- a/getting-started/elixir.html.md
+++ b/getting-started/elixir.html.md
@@ -244,7 +244,6 @@ ENV SECRET_KEY_BASE=nokey
 # Copy over the mix.exs and mix.lock files to load the dependencies. If those
 # files don't change, then we don't keep re-fetching and rebuilding the deps.
 COPY mix.exs mix.lock ./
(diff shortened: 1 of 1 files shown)


mrkurt:
This seems like a good change but I'd like @brainlid to confirm. :)

jsierles:
Thanks for the PR. Can anything in `config` affect dependency fetching?

Soon we'll be switching our guides to use `fly launch` , which will generate the current good Dockerfile for Phoenix. Once we get there we'll take this change into account.

Maxim-Filimonov:
> Thanks for the PR. Can anything in `config` affect dependency fetching?

As far as I know, it shouldn't all dependencies are specified in the `mix.exs` and `mix.lock` files

PR #21, opened 2021-09-05
fix typo in node doc

gimelg:


jsierles:
Thanks for this. These guides are being updated completely now, so will close this in favor of the new content.

PR #31, opened 2021-09-26
incorrect command in docs : flyctl postgres databases list

philberryman:
Updated to `flyctl postgres list` which seems to be the command (this is what I needed in any case).

Perhaps `flyctl postgres db list` is what some people are looking for though.


davissp14:
Thanks for catching this!

> One Postgres cluster can host multiple databases

Looks like we should be setting: `fly postgres db list <app-name>`

jsierles:
This has been resolved in another update. Thanks!

PR #38, opened 2021-12-06
Hands-on link wrong

roelandmoors:
Also mentioned here: https://community.fly.io/t/http-404-forbidden-error-from-signup-welcome-page/2881


jsierles:
Thanks!

PR #43, opened 2022-01-07
When you use ElxirLS the deployment fails by just running the getting started guide

davidvanleeuwen:
Adding a line to explain it in the troubleshooting at the end.

### DIFF
--- a/getting-started/elixir.html.md.erb
+++ b/getting-started/elixir.html.md.erb
@@ -332,3 +332,4 @@ Here are a few tips to help diagnose and identify problems.
 - Build the Dockerfile locally to verify it builds correctly. `docker build .`
 - Check the `:prod` config in `config/runtime.exs`, which is not used locally. Carefully review it.
 - Run `fly logs` to check server logs.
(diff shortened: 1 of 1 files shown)


jsierles:
Why does this cause deployment to fail?

davidvanleeuwen:
I think it's related to something else. Not sure, figured it was ElixirLS, because of:
Screenshot 2022-01-08 at 14 19 03

But it seems when I'm not running ElixirLS I'm sometimes also getting the same result:
Screenshot 2022-01-08 at 14 02 48

davidvanleeuwen:
Closing this. It's not because of ElixirLS, although it randomly pops up. I think it's due to the combination docker/apple m1/fly.

- https://community.fly.io/t/elixir-phoenix-erro-0607-cant-close-tar-writer-io-read-write-on-closed-pipe/2884/20
- https://community.fly.io/t/error-deploying-app-on-m1-pro-qemu-uncaught-target-signal-11-segmentation-fault-core-dumped/3575/3

PR #47, opened 2022-02-03
Senyo/typos

senyosimpson:
just corrections on a few typos. i also changed us mentioning in our about page about sponsoring Crystal language to sponsoring Gleam language because:
1. the link was broken
2. i'm not sure if we are still sponsoring them (not visible in our github sponsors page, though it could be through another channel)


jeromegn:
Oh right, hmm. I don't know if we're still sponsoring Hyper and Crystal.

We're sponsoring Tokio and Wireguard though.

PR #84, opened 2022-03-08
changed the reference title from 'load balancing' to 'public networking'

senyosimpson:
Currently all the details for public networking are under the title `load balancing`. I feel like it doesn't really explain the concepts. More so because the title of the content under that title is `Public Network Services`


jsierles (approved):
👍

PR #85, opened 2022-03-08
added documentation for force ssl configuration

senyosimpson:
Still need to proof-read and view locally

### DIFF
--- a/reference/configuration.html.md
+++ b/reference/configuration.html.md
@@ -203,12 +203,16 @@ For each external port you want to accept connections on, you will need a `servi
   [[services.ports]]
     handlers = ["http"]
     port = "80"
+    
+    # optional
+    http_options = { compress = false }
 ```
 
 This example defines an HTTP handler on port 80.
 
 * `handlers` : An array of strings, each string selecting a handler process to terminate the connection with at the edge. Here, the ["HTTP" handler](/docs/services/#http) will accept HTTP traffic and pass it on to the internal port of the application which we defined earlier.
 * `port` : A string which selects which external ports you want Fly to accept traffic on. You can configure an application to listen for global traffic on ports 25, 53, 80, 443, 5000, 8443, 25565, and ports 10000 - 10100.
+* `http_options` : A table that specifies configuration options for HTTP handlers. It is entirely optional and can be left out.
 
 You can have more than one `services.ports` sections. The default configuration, for example, contains two. We've already seen one above. The second one defines an external port 443 for secure connections, using the ["tls" handler](/docs/services/#tls).
 
--- a/reference/configuration.html.md
+++ b/reference/configuration.html.md
@@ -205,7 +205,7 @@ For each external port you want to accept connections on, you will need a `servi
     port = "80"
 
     # optional
-    force_ssl = true
+    force_https = true
     
     # optional
     http_options = { compress = false }
@@ -215,7 +215,7 @@ This example defines an HTTP handler on port 80.
 
 * `handlers` : An array of strings, each string selecting a handler process to terminate the connection with at the edge. Here, the ["HTTP" handler](/docs/services/#http) will accept HTTP traffic and pass it on to the internal port of the application which we defined earlier.
 * `port` : A string which selects which external ports you want Fly to accept traffic on. You can configure an application to listen for global traffic on ports 25, 53, 80, 443, 5000, 8443, 25565, and ports 10000 - 10100.
-* `force_ssl`: A boolean which determines whether to enforce HTTP to HTTPS redirects.
+* `force_https`: A boolean which determines whether to enforce HTTP to HTTPS redirects.
 * `http_options` : An optional table that specifies configuration options for HTTP handlers.
 
 You can have more than one `services.ports` sections. The default configuration, for example, contains two. We've already seen one above. The second one defines an external port 443 for secure connections, using the ["tls" handler](/docs/services/#tls).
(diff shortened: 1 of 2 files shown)


reference/services.html.md.erb
```diff
@@ -57,3 +57,111 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
...
+    [services.ports.http_options.response]
+    header = { myheader = "fly", via = false}
+
+    [services.ports.http_options.https_redirect]
+    enable = true
+    port = 777
+    status_code = 308
+```
+
+### Compress
```
jsierles, line 82: Perhaps better to name this `Disable response compression`

jsierles: Also, provide example code like all the other options do.

reference/services.html.md.erb
```diff
@@ -57,3 +57,111 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
...
+
+By default, Fly will compress response bodies. You can switch it off by setting the value to `false`.
+
+### Response
```
jsierles, line 86: Same here, maybe `Modifying Response Headers`. And below, "The `response` section allows you to add or remove HTTP headers to the response.

reference/services.html.md.erb
```diff
@@ -57,3 +57,111 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
...
+For multiple value headers
+
+```toml
+[services.ports.http_options.response]
+header = { myheader = ["fly", "io" ] }
+```
+
+#### Remove a header
+
+A header is removed by setting it to `false`
```
jsierles, line 108: Perhaps `Remove a header by setting it 'false'`

reference/services.html.md.erb
```diff
@@ -57,3 +57,111 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
...
+
+```toml
+[services.ports.http_options.response]
+header = { via = false }
+```
+
+### HTTPS redirect
+
+> This option is only applicable to HTTP handlers. If you accidentally set it on an HTTPS handler, nothing will happen
```
jsierles, line 117: Missing . here.

reference/services.html.md.erb
```diff
@@ -57,3 +57,111 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
...
+
+It is also possible to change the status code
+
+```toml
+[services.ports.http_options.https_redirect]
+enable = true
+status_code = 308
+```
+
+which will return a response like so
```
jsierles, line 160: Maybe better not to split up sentences like this. Make a full sentence here like `This configuration responds with:`

reference/configuration.html.md
```diff
@@ -203,12 +203,16 @@ For each external port you want to accept connections on, you will need a `servi
...
+    
+    # optional
+    http_options = { compress = false }
 ```
 
 This example defines an HTTP handler on port 80.
 
 * `handlers` : An array of strings, each string selecting a handler process to terminate the connection with at the edge. Here, the ["HTTP" handler](/docs/services/#http) will accept HTTP traffic and pass it on to the internal port of the application which we defined earlier.
 * `port` : A string which selects which external ports you want Fly to accept traffic on. You can configure an application to listen for global traffic on ports 25, 53, 80, 443, 5000, 8443, 25565, and ports 10000 - 10100.
+* `http_options` : A table that specifies configuration options for HTTP handlers. It is entirely optional and can be left out.
```
jsierles, line 215: Maybe just say `An optional table to specify configuration` and cull the second sentence.

reference/services.html.md.erb
```diff
@@ -57,3 +57,111 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
 
 The `proxy_proto` handler adds information about the original connection, including client IP + port and server IP + port (from the client's perspective).  Most applications need additional logic to accept the proxy protocol, either using a prebuilt library or implementing the [proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) directly.
 
+
+## _HTTP Options_
+
+`http_options` is used to specify additional configuration for HTTP handlers.
```
jsierles, line 63: Maybe activate with "Use `http_options` to specify..."

jsierles (changes requested):
Thanks for taking care of this. I suggested some changes.

reference/services.html.md.erb
```diff
@@ -57,3 +57,76 @@ If you don't specify handlers, we just forward TCP to your application as is. Th
 
 The `proxy_proto` handler adds information about the original connection, including client IP + port and server IP + port (from the client's perspective).  Most applications need additional logic to accept the proxy protocol, either using a prebuilt library or implementing the [proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) directly.
 
+## _Enforcing SSL_
+
+The `force_ssl` configuration option enforces SSL by redirecting HTTP to HTTPS. When enabled, a redirect response is returned with a `301` status code. If a TLS handler is present and this option is set, it will be ignored.
+
+At present, there is no support for redirecting to a TLS handler on a port other than 443, using a different status code or redirecting specific requests.
```
senyosimpson, line 64: @jsierles Should this be included? I'm not sure how we communicate these type of limitations

senyosimpson:
@jsierles I think we should hold off merging this in for a bit of time. At least, let me get it out on the community forum first and then we can put this in

senyosimpson:
The HTTPS redirects feature was moved out of the HTTP options. I'd rather not bundle all the changes together (and I'm not sure we want the HTTP options in the official documentation as of yet)

PR #106, opened 2022-04-06
Update overmind CMD to use Exec form

bbonamin:
overmind captures SIGTERM and cleans itself up by deleting `.overmind.sock`, but in order for this to work correctly it needs to run in Exec form, which is the recommended by docker (https://docs.docker.com/engine/reference/builder/#cmd)


jsierles:
Nice catch!

PR #109, opened 2022-04-08
Add warning about ENTRYPOINT and CMD

sashaafm:
Add a warning about `ENTRYPOINT` and `CMD` failing to work in a `Dockerfile` and should be done in the `fly.toml` instead.

This PR warns about what happened in this [community thread](https://community.fly.io/t/deployment-always-fails-and-container-logs-are-not-printed/4631).


jsierles:
Thanks for the PR. `ENTRYPOINT` and `CMD` work on Fly, so what's written in forum (and this PR) is not accurate.

Migrations should always be handled via `release_command` because if you run it in an ENTRYPOINT, every VM that boots up will end up running migrations. You want migrations to happen _once_, and usually, for a deployment to fail should the migration fail. That's what `release_command` is for.

PR #130, opened 2022-04-28
Add all available Fly-exposed metrics to metrics docs

jsierles:
This PR should also demystify the units on ambiguous values like those named `*disk_time*`

### DIFF
--- a/reference/metrics.html.md
+++ b/reference/metrics.html.md
@@ -96,27 +96,84 @@ fly_edge_data_out
 fly_edge_data_in
 ```
 
-### Instance series
+## Instance series
 
-These are pulled and exposed from inside your app instances:
+These are pulled and exposed from your app VMs.
+
+### Instance memory
+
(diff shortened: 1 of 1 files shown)


jeromegn:
I believe the disk metrics should work for all disks: the root and all mounted volumes.

jeromegn:
The `fly_instance_disk_*` metrics should have all disk from the VM, but they're not directly linked to Fly volumes. You should be able to see `vda`, `vdb`, etc. if you have multiple disks.

jsierles:
Cool, I'll give that a try. I suppose for now it's not a problem that you can't link it to a volume name.

jeromegn:
> Cool, I'll give that a try. I suppose for now it's not a problem that you can't link it to a volume name.

Oh right. Sorry, I was only commenting on this line really: `### Instance Root Disk`. It should read `### Instance Disks` or something like that because it's not just the root disk.

PR #164, opened 2022-05-28
Change the case of 'IP' in `Fly-Client-Ip`

omern1:
The `Fly-Client-IP` header is spelled as `Fly-Client-Ip` by the runtime environment.

`Fly-Client-Ip:[92.xxx.xxx.xx]`

### DIFF
--- a/reference/runtime-environment.html.md
+++ b/reference/runtime-environment.html.md
@@ -26,7 +26,7 @@ Request headers carry information that is specific to the incoming request and i
 
 ## _Request Headers_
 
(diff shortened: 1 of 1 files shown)


mrkurt:
Which runtime? Runtimes normalize header names differently.

omern1:
Sorry I'm not very familiar with the how Fly.io works, I was referring to the runtime metioned here: https://fly.io/docs/reference/runtime-environment. It may be the Go runtime in my case but I don't think it fiddles with headers.

PR #165, opened 2022-06-02
Update custom-domains-with-fly.html.md

jamilabreu:
Make options skimmable, and instructions more specific


jsierles:
Thanks, this is helpful.

PR #176, opened 2022-06-13
Add `ELIXIR_ERL_OPTIONS` to `rel/env.sh.eex` file

derrickreimer:
The default `env.sh.eex` file generated by `mix release.init` does not include the `ELIXIR_ERL_OPTIONS` export:

https://github.com/elixir-lang/elixir/blob/7e4fbe657dbf9c3e19e3d2bd6c17cc6d724b4710/lib/mix/lib/mix/tasks/release.init.ex#L59-L78

However, it appears this line is necessary in order for Fly networking to work properly (see https://community.fly.io/t/could-not-contact-remote-node-reason-nodedown-aborting/5221/3?u=derrickreimer).

Relatedly, I was not able to successfully start an IEx session (following [these instructions](https://fly.io/docs/getting-started/elixir/#iex-shell-into-your-running-app)) _until_ I made these updates to the `env.sh.eex` file. The instructions for changing `env.sh.eex` are listed _below_ the IEx instructions, so it might be trip-up point if folks are following this guide sequentially. I would recommend moving this portion higher in the guide, but have refrained from doing so in this PR, since that's a bigger structural change.

Thanks all!

### DIFF
--- a/getting-started/elixir.html.md.erb
+++ b/getting-started/elixir.html.md.erb
@@ -143,6 +143,7 @@ Then edit the generated `rel/env.sh.eex` file and add the following lines:
 ip=$(grep fly-local-6pn /etc/hosts | cut -f 1)
 export RELEASE_DISTRIBUTION=name
 export RELEASE_NODE=$FLY_APP_NAME@$ip
+export ELIXIR_ERL_OPTIONS="-proto_dist inet6_tcp"
 ```
 
 This names our Elixir node using the Fly application name and the internal IPv6 address. Make sure to deploy after making this change!


brainlid:
Hi @derrickreimer!

I created a new Phoenix app (v1.6.10) and following the guide I had no problems. 

I got an IEx shell to the newly deployed app.

I then went ahead and ran the `mix release.init`. After doing that and deploying I was still able to IEx into the node.

The ENV setting is appended to the Dockerfile by the `flyctl` program.

```dockerfile
# Appended by flyctl
ENV ECTO_IPV6 true
ENV ERL_AFLAGS "-proto_dist inet6_tcp"
```
It sets it under a different key, but it works for me.

Are you working with a custom Dockerfile that doesn't have this setting or was your Phoenix app generated with an older version of Phoenix?

We used to have that setting in the documentation but IIRC, it was removed because it can cause problems in other scenarios? Not sure right now.

derrickreimer:
Ah, interesting! I used the Dockerfile generated by `mix phx.gen.release --docker` and I'm not sure why those `ENV`s did not get appended when I deployed to Fly initially. I believe I had already created a `fly.toml` file, so perhaps those auto-scaffolding steps were skipped?

At any rate, I added those appends to my Dockerfile (and removed `ELIXIR_ERL_OPTIONS` from `env.sh` and it works! This might be enough of an edge case that documentation about it is not warranted, but could be helpful to include in a troubleshooting section somewhere for those who are retrofitting an existing Dockerfile.

Thanks for the help!

brainlid:
Thanks for the follow-up! I agree that it should be included as a troubleshooting step if not elsewhere as well.

Thanks!

brainlid:
@derrickreimer I updated the docs to hopefully help future devs who hit this issue. 

You can checkout the update here:
https://fly.io/docs/getting-started/elixir/#important-ipv6-settings

Let me know if you feel it could be improved and how. 

Thanks!

derrickreimer:
That looks great!

PR #183, opened 2022-06-23
UDP works in port-forwarding mode only

ignoramous:
Ref: https://community.fly.io/t/unexpected-udp-port-exposing-behaviour/5736

cc: @tqbf


ignoramous:
https://github.com/superfly/docs/commit/8f11c25b474ac2a46f851399fdb1330c89e61ba8

PR #186, opened 2022-06-25
Cross-app fly-replay guide

catflydotio:
fly-proxy has new abilities we can access with the fly-replay header, notably the ability to route the replayed request to another app within the same organization. This allows a convenient pattern for developers building their own cloud for their customers. We should have some sort of guide showing what you can do with it.

This PR should be finalized after (or at least not before) https://github.com/superfly/docs/pull/185, which adds a reference entry for fly-replay.


catflydotio:
This is a rough start. Any input welcome.

kcmartin:
Closing out older PRs and drafts

PR #188, opened 2022-06-27
add MacPorts as alternative to Homebrew; add missing Homebrew links

tsujp:
I was going through the guide to using Fly and the documentation only mentions Homebrew when there are two package managers for macOS: Homebrew, and MacPorts. I think it's important people at least know MacPorts is an option here.

MacPorts does package and keep flyctl up to date.


jsierles (approved):

andie787:
Stale PR since updates weren't made. Can create a new PR if needed.

PR #202, opened 2022-07-11
Fix sentence in private networking docs

fideloper:
Does this make sense?

### DIFF
--- a/reference/private-networking.html.md
+++ b/reference/private-networking.html.md
@@ -92,7 +92,7 @@ Examples of retrieving this information are in the [fly-examples/privatenet](htt
 
 ## Private Network VPN
 
-In conjunctin with our [6PN private networking](/docs/reference/private-networking/), we can use [WireGuard](https://wireguard.com/) VPN software. This is a flexible and secure way to plug into each one of your Fly organizations and connect to any and all apps within that organization.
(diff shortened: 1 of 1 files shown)


reference/private-networking.html.md
```diff
@@ -92,7 +92,7 @@ Examples of retrieving this information are in the [fly-examples/privatenet](htt
 
 ## Private Network VPN
 
-Along with our [6PN private networking](/docs/reference/private-networking/),  [WireGuard](https://wireguard.com/) networking software. This is a flexible and secure way to plug into each one of your Fly organizations and connect to any and all apps within that organization.
+In conjunctin with our [6PN private networking](/docs/reference/private-networking/), we can use [WireGuard](https://wireguard.com/) VPN software. This is a flexible and secure way to plug into each one of your Fly organizations and connect to any and all apps within that organization.
```
catflydotio, line 95: There's a typo in "conjunction"--but would it be easier if we just say "You can use [WireGuard](https://wireguard.com/) VPN software to connect to our [6PN private network](https://fly.io/docs/reference/private-networking/) from outside."?

fideloper: ooph and also yes, agree!

PR #219, opened 2022-07-29
Support tweaks

eliferrous:
provided a bit more detail for how ssh root certs work. 
also made an unrelated change to `fly orgs delete` to prevent common mistakes


catflydotio:
These particular doc files get generated using flyctl source. For instance, https://github.com/superfly/flyctl/blob/master/internal/command/orgs/delete.go contains a string `long` with the value 
```
`Delete an existing organization.
`
```

The "see also"s are generated from the directory structure, I believe, so it's harder to tweak those.

PR #238, opened 2022-08-06
Update flag for rails console command

afomera:
I spent a few minutes trying to connect with -c when I really needed -C instead for the Rails console.

--

I'm unsure if this is worthy of an issue / doc fix... but I also had to prefix my commands with the WORKDIR from my Dockerfile that was generated by `fly launch` (app/) to make the commands work... so `bin/rails console` (doesn't work) -> `app/bin/rails console` (works)


bradgessler:
Oh dang, I wrote those docs. Sorry for wasting 5 min of your life. Feel good knowing that you're collectively saving several hours of many other peoples' lives—its definitely worthy of an issue to fix!

PR #240, opened 2022-08-06
Updated docs for flyctl suspend & resume to match the deprecation notice from flyctl

GarrettCannon:
Not sure if the styling I picked is okay, please let me know if there are any tweaks needed


jsierles:
Thanks for the PR. Unfortunately, these changes would be overwritten by automated CLI docs updates. I believe we can make this work directly in flyctl, though. It may be worth checking there if nobody else gets to it.

PR #243, opened 2022-08-09
A doc on "how app launching works" and some nav changes to fit it

catflydotio:
image

I can break this up, or change it, but basically I made some tentative nav changes to make it easier to find docs for fundamental things, then added a doc for a fundamental thing (launching a new app). The main nav still needs changes, I think, but that can be done progressively.

So far it's pretty non-destructive: the only URLs that have changed are inside the Hands-on. I did add redirect entries for these but can't test them from Middleman.

Moving the content of the Hands-on into partials sets the stage for improving the instructions for each step and using it both as a standalone reference and as part of a wizardy guided flow within Hands-on. If this turns out not to be a good pattern it's easy to undo or ignore, but it does already clean up the "install flyctl" doc which was three (3) almost-identical files. This way we only need to maintain one.


bradgessler:
> I did add redirect entries for these but can't test them from Middleman

I can try hacking Middleman to make this work, but when I did research for this a few days ago I ended up on a GH issue I opened a few years ago at https://github.com/middleman/middleman/issues/1831 😆, which reminded me of why I made my own site generator. I hit my time box and moved on, but I didn't have as tangible of a need to get it working as I think you do now.

Anyway, what I can do pretty quickly is add a banner to the redirect page (or something) that says "this page has been moved to __________" in the view of the existing page, which would have to be kept there for the time being.

Another thing that hasn't been implemented yet is the actual redirector in production that reads the `/redirects.json` file.

catflydotio:
I added the doc and separated Getting Started from the Quickstart Guides in nav, but didn't change any existing URLs. Keeping this branch for the moment and may return to the other aspects of this after redirects are sorted out.

PR #244, opened 2022-08-09
Update configuration.html.md

rubys:
Expand on the limitations of what can be done during a deploy step, and suggest an alternative.

### DIFF
--- a/reference/configuration.html.md
+++ b/reference/configuration.html.md
@@ -120,9 +120,9 @@ This section configures deployment-related settings such as the release command 
   release_command = "bundle exec rails db:migrate"
 ```
 
-This command runs in a temporary VM - using the successfully built release - *before* that release is deployed. This is useful for running database migrations.
+This command runs in a temporary VM - using the successfully built release - *before* that release is deployed. This is useful for running Postgres database migrations.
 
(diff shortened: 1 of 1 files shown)


reference/configuration.html.md
```diff
@@ -120,9 +120,9 @@ This section configures deployment-related settings such as the release command
   release_command = "bundle exec rails db:migrate"
 ```
 
-This command runs in a temporary VM - using the successfully built release - *before* that release is deployed. This is useful for running database migrations.
+This command runs in a temporary VM - using the successfully built release - *before* that release is deployed. This is useful for running Postgres database migrations.
```
catflydotio, line 123: Here's a chance for me to learn something. Would a release command not be useful for any database not using your app's mounted volume (i.e. not-SQLite)? Or is this really Postgres-specific?

rubys: The release command would be useful for all sorts of things involving network accessible data.  Just not for data on this VM's image or any persistent volumes.

One could argue that the sentence could be struck completely, but this would leave working out what this feature could be useful for as an exercise for the student.

My feeling is leaving the sentence in as an example of what this feature can be used for is valuable, even if the sentence doesn't aim to exhaustively enumerate all the potential uses for this feature.

PR #248, opened 2022-08-10
Fix 404ing link

natematykiewicz:
On https://fly.io/docs/rails/the-basics/sidekiq/, the "outlined in the multiple-proccesses" link returns a Nginx 404 page when you click it. It needs to be https://fly.io/docs/app-guides/multiple-processes not https://fly.io/docs/app-guides/multiple-processes.html.md.


bradgessler:
Doh thanks for catching this!

PR #266, opened 2022-08-17
Updated Redis documentation

enesakar:
- added `Connect to an Upstash Redis database` section
- some other minor updates


jsierles (approved):
Thanks. Will do a few updates after merging.

PR #278, opened 2022-08-23
Add service.concurrency.type documentation for fly.toml

tvdfly:
Also, include a new page on load balancing 🎉

### DIFF
--- a/reference/load-balancing.html.md
+++ b/reference/load-balancing.html.md
@@ -36,7 +36,7 @@ Closeness is determined by RTT (round-trip time) between the Fly.io edge node re
 
 ## Examples
 
-To help clarify how the Fly.io platform makes load balancing decision, we provide some example configuration and scenarios:
+To help clarify how the Fly.io platform makes load balancing decisions, we provide some example configuration and scenarios:
 
 ### Web service
 
--- a/reference/load-balancing.html.md
+++ b/reference/load-balancing.html.md
@@ -1,0 +1,73 @@
+---
+title: Load Balancing
+layout: docs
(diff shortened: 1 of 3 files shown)


michaeldwan:
Oh nice, several customers were tripped up recently because they didn't know about the `type` setting. 

Consider updating the `concurrency` section to explain load balancing instead of autoscaling since that's the most important use for those settings now (eg debugging in #glideapps). Our autoscaling with nomad is slow and unreliable... fine to ignore until we have it for machines. There's a ton of conversations about load balancing on the forums and slack. Here's a good recent explanation: https://community.fly.io/t/load-balancing-within-a-region/5962.

tvdfly:
Thanks @michaeldwan ! I added a page about load balancing and linked to it from the configuration docs.

reference/load-balancing.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Closeness
+
+Closeness is determined by RTT (round-trip time) between the Fly.io edge node receiving a connection or request, and the worker node where your instance runs. Even within the same region, we use different datacenters with different RTTs. These RTTs are measured constantly between all servers.
+
+*Note:* These values are not visible to Fly.io users. We share this information here to help clarify how load balancing decisions are made.
+
+## Examples
+
+To help clarify how the Fly.io platform makes load balancing decision, we provide some example configuration and scenarios:
```
catflydotio, line 39:
```suggestion
To help clarify how the Fly.io platform makes load balancing decisions, we provide some example configuration and scenarios:
```

PR #281, opened 2022-08-25
Update the Set API key section

bryanhonof:
I'm not aware of a `fly` command, but `flyctl` provides the same `auth token` sub-commands. So I am assuming it should be `flyctl` instead of `fly`? Also, clarify that you _need_ to be logged in before you can request a token.


fideloper:
This is out of date now, but if there's still an issue (or if you feel this would still be useful) you can pull down latest and add it!

PR #285, opened 2022-08-26
grammar correction to log-in-to-fly.html.md

studiozandra:
I really just came to fix **'a details'**, but then noticed a couple of other things.
'sign in' for verb form, 'sign-in' for adjective form; 2 clauses separate with semicolon


bradgessler:
Thanks!

PR #313, opened 2022-09-04
Remove single quotes when migrating Heroku config

bensheldon:
Heroku will wrap some configuration values with single quotes; this uses `sed` to remove those quotes.


skatkov:
Had a similar issue, great addition!

fideloper:
Any idea if this is compatible across MacOS and Linux? IIRC the `sed`'s are different but that might just be when doing a replace on a file.

kcmartin:
Thanks for this, and sorry for the long delay! This guide moved to `getting-started/migrate-from-heroku.html.md`, which already strips Heroku's wrapping quotes via `sed`. Closing since it's now covered. Appreciate the contribution!

PR #319, opened 2022-09-07
Update and simplify EdgeDB guide

colinhacks:
Supersedes #288


colinhacks:
Overrides #288

colinhacks:
Sorry for the ping @benbjohnson. We're getting feedback from users that the current guide isn't working due to the recently API changes in `flyctl postgres connect`. I'm the original author of this guide and I've thoroughly tested the modified version - could we get a review when you have a chance?

PR #325, opened 2022-09-09
doc: correct migration from heroku postgres to assure data is created incorrect db

daniel-sc:
Note: `-C` (`--create`) is _not_ `-c` (`--clean`). The former leads to the creation of a db with the name of the db in heroku, but we need to copy the data into the db of the given `$DATABASE_URL`.


fideloper:
I think these docs were updated and made this PR obsolete (it appears that clean is used on purpose). Let us know if that seems incorrect still.

PR #330, opened 2022-09-15
Linux can also use Homebrew

Red54:


fideloper:
We prefer Linux users use our installer over `brew`, although those who prefer/want to use Brew in their linux install are free to do so.

PR #335, opened 2022-09-19
command with "fly ..." become "flyctl ..."

olivsinz:
Here's what I'm getting when using "fly" command.

fly: The term 'fly' is not recognized as a name of a cmdlet, function, script file, or executable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again


catflydotio:
Hi @olivsinz, both `fly` and `flyctl` should work. If your system can't find the `fly` command, something's not right. You can find help with this on our community forum [https://community.fly.io](https://community.fly.io).

olivsinz:
Thanks @catflydotio I will figure it out.

PR #338, opened 2022-09-21
Add reference to full stack laravel post in the run a laravel app doc…

KTanAug21:
… to point readers to a complete guide on configuring their laravel app with basic services


fideloper:
Looks good! Merge it in whenever you'd like!

PR #364, opened 2022-10-07
Add Postgres section in Laravel Databases, add Start Script option in…

KTanAug21:
… database migration list

### DIFF
--- a/laravel/the-basics/databases.html.md
+++ b/laravel/the-basics/databases.html.md
@@ -180,6 +180,173 @@ fly secrets set DB_USERNAME=<DB_USERNAME> DB_PASSWORD=<DB_PASSWORD>
 fly deploy 
 ```
 
+## _Postgres in Fly.io_
+We have a <i>[whole section](/docs/postgres/)</i> dedicated for [Postgres](https://www.postgresql.org/), we even have a `flyctl` integration to simplify the creation and management of your Postgres Fly Apps!
+
+In this section you'll get to quickly:
+1. Setup a Postgres Fly App
+2. Connect From a Laravel Fly App by Attachment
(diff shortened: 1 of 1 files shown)


fideloper (approved):

laravel/the-basics/databases.html.md
```diff
@@ -180,6 +180,178 @@ fly secrets set DB_USERNAME=<DB_USERNAME> DB_PASSWORD=<DB_PASSWORD>
...
+```
+<small>In case you've forgotten about your app name, but managed to save the output from the Postgres Fly App creation (which you should've!), then check the value of the `Hostname`. The App name would simply be every little character before ".internal".</small> 
+
+
+2) Great, you're connected! Go ahead and complete your setup with a new database:
+```postgres
+create database <DB_NAME>;
+```
+
+3) Now if you check your database list, you should find the recent database created:
```
rugwirobaker, line 238: I don't if you this but there is `fly postgres db list` command that runs this query for you.

KTanAug21: *--super-excited typing--->*

Hi @rugwirobaker! It's awesome to have you in the Laravel Docs section! 

Now, for the step 3 above, it comes from a walkthrough of connecting with the Postgres Fly App through the `fly postgres connect -a <appname>" command. My idea was to show the reader the option of how they can connect to the Postgres Fly App from their console _then_  create a db from there, and of course while they are at it, why not list the db they've created from the console.

But it's great you've pointed out that there is a separate command for listing the Postgres Fly App's list of db using `fly postgres db list -a <appname>`, it's a great addition to the list of ways/approaches to interact with the postgres fly app, I'll add that command in there  in my next commit, and I'll be sure to add you in as a reviewer!

PR #369, opened 2022-10-12
[WIP] Django docs first commit

wsvincent:


jsierles:
Thanks for the contribution. One thing to note is that `flyctl` should already be detecting Django apps and planting a `Dockerfile`. Could you test that out, and, if it works as you expect, update these docs? Also, if anything doesn't make sense with the launcher, do open an issue on `flyctl` about it.

wsvincent:
Sweet! One thing though is that it seems `cmd` automatically adds a `$` prompt at the beginning of lines. I didn't realize that because I can't preview the files locally. But the `$` prompt should come **after** the virtual environment `(.venv)`.

Under _Getting Started_ it should be:
- `(.venv) $ django-admin startproject demo .` not `$ (.venv) $ django-admin startproject demo .`
- `(.venv) $ python -m pip install gunicorn`
- `(.venv) $ pip freeze > requirements.txt`
- `(.venv) $ fly launch`
- `(.venv) $ fly deploy`
- `(.venv) $ fly open`

Same under _Existing Django Apps_ as well for:
- `(.venv) $ fly launch`
- `(.venv) $ fly deploy`
- `(.venv) $ fly open`
- `(.venv) $ fly logs`
- `(.venv) $ fly ssh console`
- `(.venv) $ `fly ssh console -C 'python /code/manage.py createsuperuser'` And this doesn't need the `` ticks around it either.

What's the best way to fix this?

PR #378, opened 2022-10-13
Fix broken link to "The Basics"

elliotrushton:
The last sentence on this page says `it is time to move on to [The Basics](https://fly.io/docs/the-basics/).` but that link 404's.

However on the rendered page just below it there's a list of "Additional Resources" which includes a link to `[The Basics](https://fly.io/docs/rails/the-basics/)` which works -- so this PR makes the broken link match the working link.


bradgessler:
Thanks!

PR #380, opened 2022-10-14
Ignore .DB_Store files

mbuffa:
:wave:


bradgessler:
Thanks! Very annoying that macOS still does this.

PR #385, opened 2022-10-19
adding spaces to litefs.yml consul section

zaherg:
The current docs show the items at the same level as `consul` and when they are copied they are copied as is, so its `yml` and spaces are important if we really want the configuration to work

this is how the old page looks like

image

and this is how it should look like

image


zaherg:
never mind, I don't think this will change anything

PR #392, opened 2022-10-24
Fix links to postgres doc

norbajunior:
The links to Postgres doc were pointing to an inexistent file.


catflydotio (approved):
Thanks for the heads-up, @norbajunior.

PR #397, opened 2022-10-27
Fixed link for "expanded to more regions"

oliver-kriska:


bradgessler:
Thanks! I thought I found all of these 😄

PR #412, opened 2022-11-03
Explain port configuration for NextJS apps

benjreinhart:
Ran into the "unhealthy allocations" error today when launching a new NextJS app. I was following this guide and it didn't mention the port needing to be `8080`. I imagine this will be common source of confusion (e.g., [this thread](https://community.fly.io/t/source-of-deployment-failing-due-to-unhealthy-allocations/4247)) given NextJS will launch on port `3000` by default.

### DIFF
--- a/languages-and-frameworks/nextjs.html.md.erb
+++ b/languages-and-frameworks/nextjs.html.md.erb
@@ -16,6 +16,18 @@ If you just want to see how Fly deployment works, follow these steps.
 
 <%= partial "partials/flyctl" %>
 
+For your NextJS app to properly boot, it needs to be listening on the same port as `internal_port` in your `fly.toml`, port `8080`. In `package.json`, edit the `start` script:
+
+```diff
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+-   "start": "next start",
++   "start": "next start --port 8080",
+    "lint": "next lint"
+  },
+```
+
 Now let's launch your NextJS app.
 
 <%= partial "partials/launch", locals: { detected: "NextJS", app_name: app_name } %>
@@ -26,7 +38,7 @@ Now let's launch your NextJS app.
 
 If you're an NextJS user you might know that it supports [exposing environment variables to the browser](https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser) using variables with name starting with `NEXT_PUBLIC_`.
 
-For our build system to understand that you need to tweak two sections we generate for you, `fly.toml` and our `Dockerfile`. 
+For our build system to understand that you need to tweak two sections we generate for you, `fly.toml` and our `Dockerfile`.
 
 Search for `[build.args]` on your `fly.toml` and add the variables you need there.
 


lubien:
Hi @benjreinhart thanks for taking the time to contribute.

We should set on your `fly.toml` `[env]` section `PORT=8080` which would be enough to change the port for that, didn't that work when you `fly launch`-ed your app?

benjreinhart:
The `fly launch` command created the following `fly.toml` (which I did not edit):

```
# fly.toml file generated for privy-test on 2022-11-03T12:57:44-04:00

app = <name>
kill_signal = "SIGINT"
kill_timeout = 5
processes = []

[env]

[experimental]
  allowed_public_ports = []
  auto_rollback = true

[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["app"]
  protocol = "tcp"
  script_checks = []
  [services.concurrency]
    hard_limit = 25
    soft_limit = 20
    type = "connections"

  [[services.ports]]
    force_https = true
    handlers = ["http"]
    port = 80

  [[services.ports]]
    handlers = ["tls", "http"]
    port = 443

  [[services.tcp_checks]]
    grace_period = "1s"
    interval = "15s"
    restart_limit = 0
    timeout = "2s"
```

When Fly booted my next js app, it did so running `next start` which boots up the Next app on localhost:3000. Confirmed this by looking at logs via `fly logs`:

```
2022-11-03T16:59:19Z app[] ewr [info]$ next start
2022-11-03T16:59:19Z app[] ewr [info]ready - started server on 0.0.0.0:3000, url: http://localhost:3000
```

 At that point, flyctl hung for a while before finally crashing with the unhealthy allocations error.

Here is another thread I found with this error: https://community.fly.io/t/failed-due-to-unhealthy-allocations/199

If this solution is not the right one, I'm happy to update it. But in either case, it seems like some extra configuration was needed and that was missing from the guide.

lubien:
Would you happen to have a Dockerfile before running `fly launch`? Did `fly launch` say something like "Detected NextJS app"?

https://github.com/superfly/flyctl/blob/master/scanner/nextjs.go#L9

We recently fixed this and added the env bit but your `fly.toml` doesn't have it which is weird

benjreinhart:
I don't remember if it said that now. `fly launch` did create the dockerfile and I did not edit it. I'm using:

```
flyctl v0.0.427 darwin/arm64 Commit: 72736560 BuildDate: 2022-11-02T20:17:21Z
```

benjreinhart:
Agree the better solution is definitely to have it automated rather than more work on end user's part! So happy to close this out if that's what is supposed to be happening.

lubien:
Really appreciate you taking your time on reporting and trying to fix this ❤️

lubien:
```
nextjs-blog on  main [?] via ⬢ v16.11.1
[I] ➜ fly launch
Creating app in /Users/lubien/workspace/fly-tests/nextjs-blog
Scanning source code
Detected a NextJS app
? Overwrite "/Users/lubien/workspace/fly-tests/nextjs-blog/.dockerignore"? Yes
? Choose an app name (leave blank to generate one):
? Select Organization: Lubien (personal)
? Choose a region for deployment: São Paulo (gru)
Created app throbbing-haze-344 in organization personal
Wrote config file fly.toml
? Would you like to deploy now? Yes
```

I've just run a clean launch and it seems ok on the latest flyctl

benjreinhart:
Cool... I manually edited `fly.toml` with `PORT = 8080` under `[env]` and removed the port override in the start script and it works fine. But not sure why it didn't do that automatically if it's supposed to

PR #414, opened 2022-11-04
Add note about Procfile overwriting to python.html.md.erb

jon4syth:


brainlid (approved):
Thanks Jon for helping to improve the reader's experience!

PR #420, opened 2022-11-09
Update speedrun.html.md.erb

mishrasur7:
I am quite new to Fly.io. In case if users no longer need hosting in Fly.io, what are the steps?. I cannot find it in the docs, I guess its worth to mention that as well in the docs.


catflydotio:
Hi @mishrasur7, you can open an issue for this if you like. It's true that this would be a good addition to docs. 

But if you need information that's currently not in docs, your best bet is to head over to the forum at https://community.fly.io.

PR #421, opened 2022-11-09
Update broken link on migrate-from-heroku.html.md

Keo222:
On the very last line of this file there is a link leading to "/docs/postgres/this-basics/connecting". This link is broken. I believe the link should actually be to "/docs/postgres/the-basics/connecting". Changing the word "this" to "the" within the link.


bradgessler:
Thanks so much!

PR #423, opened 2022-11-10
Update getting-started.html.md

johnwesonga:
When building the docker image this error **"failed to solve with frontend dockerfile.v0: failed to create LLB definition"** keeps popping up. Fix for this is to type in the terminal:

```
export DOCKER_BUILDKIT=0
export COMPOSE_DOCKER_CLI_BUILD=0
```

before the fly launch command

### DIFF
--- a/litefs/getting-started.html.md
+++ b/litefs/getting-started.html.md
@@ -67,6 +67,13 @@ configure your application so it's publicly visible over HTTPS.
 
 ### Launching your app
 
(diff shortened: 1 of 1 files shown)


fideloper:
This seems like an issue that might be specific to your use case, I don't believe these should be necessary. If you'd like to post in the [community forum](https://community.fly.io/), maybe we can sus out what's going on.

PR #445, opened 2022-11-22
Quote environment variable in Markdown

gma:
Without the backticks around the variable name the "_DATABASE_" text is converted to italics.


catflydotio:
Cheers @gma

PR #456, opened 2022-11-27
Fixing postgres version number

brycethornton:
It appears the version of Postgres listed in the docs was out-of-date.


catflydotio:
Thanks @brycethornton. Closing just because this was fixed incidentally in another commit.

PR #483, opened 2022-12-15
Change two links pointing to localhost:4567 to be relative instead

Jesse-Bakker:
These links are not resolvable in the hosted version of the docs


fideloper:
thank you!

PR #499, opened 2022-12-31
Fix 404 and remove useless Additional

selfsryo:
- On https://fly.io/docs/django/, `Existing Django Apps` link returns 404 page when you click it. I fixed it
- On https://fly.io/docs/django/getting-started/existing/,  `Existing Django Apps` link in `Additional resources` returns same page. That's why I removed it


andie787:
Thanks! Link was fixed in a previous PR and #499 fixes the self-link.

PR #501, opened 2023-01-01
Fixed link to install flyctl

codepope:
Noticed in passing... was linking to non-existant getting-started/install...


catflydotio:
Cheers @codepope

PR #507, opened 2023-01-06
Update documentation with correct commads

Anshuman71:
Replace `fly` with `flyctl` in command line code blocks.


catflydotio:
`flyctl` and `fly` will both work but we're preferring `fly` in docs. (That's a work in progress.)

PR #510, opened 2023-01-07
Differentiate Fly.io database projects from examples

catflydotio:
To make it clearer that Fly Postgres, LiteFS, and Upstash are special and everything else is generic

### DIFF
--- a/database-storage-guides.html.md
+++ b/database-storage-guides.html.md
@@ -1,7 +1,7 @@
 ---
 title: Databases & Storage
 layout: docs
-toc: false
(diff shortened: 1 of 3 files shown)


database-storage-guides.html.md
```diff
@@ -1,24 +1,50 @@
 ---
-title: Database and Storage Guides
+title: Databases & Storage
 layout: docs
-toc: true
+toc: false
```
catflydotio, line 4:
```suggestion
toc: true
```

PR #513, opened 2023-01-10
Document fly-prefer-region, combined with fly-replay in a general request routing section

jsierles:


reference/fly-replay.html.md
```diff
@@ -1,60 +1,90 @@
...
+Clients accessing Fly.io apps may append the `fly-prefer-region` header to attempt sending the request directly to a desired target region. This is useful for cases where `fly-replay` isn't practical, such as large file uploads which can't be replayed once buffered by the proxy.
+
+If the target regions hosts no healthy instances, the nearest region with healthy instances will field the request.
+
+Example:
+```
+fly-prefer-region: ams
+```
 
 ## Example use cases
```
catflydotio, line 48: Tempted to move these under fly-replay since they're all fly-replay examples.

reference/fly-replay.html.md
```diff
@@ -1,60 +1,90 @@
...
+Replay a request to the region where the HA database leader lives:
 ```
 fly-replay: region=sjc
-``` 
+```
+
 The proxy will get the request to an instance in that region and let the instances in the cluster take care of getting it to the writeable leader.
 
-Send the request to an instance of a different app in the same organization, in the nearest region that has one:
+### Replaying requests to other apps
```
catflydotio, line 62:
```suggestion
### Replay requests to other apps
```

reference/fly-replay.html.md
```diff
@@ -1,60 +1,90 @@
...
+`fly-replay` can replay requests across apps in the same organization. Think of a router app for a FaaS that wants to spin up a customer [VM](/docs/reference/machines/) on demand.
+
+To send the request to an instance of a different app in the same organization, in the nearest region that has one:
 ```
 fly-replay: app=app-in-same-org
 ```
-This can be used for cross-app replays; think a router app for a FaaS that wants to spin up a customer [VM](/docs/reference/machines/) on demand.
 
-Replay the request to a specific instance by ID:  
+### Replaying requests to specific VMs
```
catflydotio, line 71:
```suggestion
### Replay requests to specific VMs
```

reference/fly-replay.html.md
```diff
@@ -1,60 +1,90 @@
...
+
+Replay the request to a specific VM by ID:
 ```
 fly-replay: instance=00bb33ff
 ```
 
-Tell the target instance why the request was routed to it (where the recipient app has logic to make use of that information):
+### Passing state to replay targets
```
catflydotio, line 78:
```suggestion
### Pass state to replay targets
```

reference/fly-replay.html.md
jsierles: Yeah probably a good idea.

PR #517, opened 2023-01-10
add documentation for external port ranges / flyctl command

lillianberryfly:
unsure on phrasing. I'm open to suggestions 😄


catflydotio:
Seems good to start with! Only comment: the flyctl command docs are generated from flyctl source.

PR #521, opened 2023-01-11
docs: update machines docs for syntax highlighting and fix parsing issues

scarlson:


jeregrine:
@scarlson first of all thank you so much for this contribution!

When I test this locally it highlights some problems with our current code highlight tool and those escaped underscores are absolutely necessary. If you could re-submit your PR with just the ````sh` I can merge it right away but I spent a bunch of time trying to find an easy workout and failed.

jeregrine:
@scarlson I am being told that's not also in the cards because well:

image

There is an internal project underway to modernize the static site. For now I need to close this.

Issue #522, opened 2023-01-12
Edit this page on GitHub may not help the reader fix an issue


SvenDowideit:
on https://fly.io/docs/hands-on/launch-app/ the `Edit` link takes me to https://github.com/superfly/docs/edit/main/hands-on/launch-app.html.md.erb which contains none of the text I was reading.

jeregrine:
@SvenDowideit thanks for pointing this out!

PR #525, opened 2023-01-12
Removing unnecessary hands-on partials as they are unused anywhere else.

jeregrine:
Thanks to #522 for pointing this out.


bradgessler:
I just want to add, removing these partials is insanely helpful for a migration project I'm working on. Thanks!

PR #531, opened 2023-01-13
Some text updates (based on feedback in the forum)

gregmsanderson:
Someone in the community forum pointed out some problems they had based on the guide. I fixed a few issues they mention (e.g there's no .gitignore in repo, it's flyctl launch that makes the fly.yaml, sample go repo uses a master branch rather than main).

The YAML formatting is also slightly wrong on the live docs page. However it looks ok in the github editor/preview so the markdown is correct ... hmm. Perhaps it wants tabs? Not sure.

### DIFF
--- a/app-guides/continuous-deployment-with-github-actions.html.md
+++ b/app-guides/continuous-deployment-with-github-actions.html.md
@@ -9,7 +9,7 @@ categories:
   - github
   - guide
 priority: 10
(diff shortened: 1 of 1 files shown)


andie787:
Most of this (the important bits) was solved by https://github.com/superfly/docs/pull/751.

PR #536, opened 2023-01-16
Escape unescaped underscores in Machines leases partials

isms:
Unescaped underscores were turning into italics in several curl literal blocks here: https://fly.io/docs/machines/working-with-machines/

image


andie787:
Fixed by #941

PR #539, opened 2023-01-17
Laravel volumized storage section

KTanAug21:
Add section for mounting volume to the storage folder, which requires reinitialization after volume mount

### DIFF
--- a/laravel/partials/_volume_storage_steps.html.md.erb
+++ b/laravel/partials/_volume_storage_steps.html.md.erb
@@ -15,7 +15,6 @@
       source="storage_vol"
       destination="/var/www/html/storage"
   ```
-    
 
     <%= partial "docs/laravel/partials/volume_storage_warning" %>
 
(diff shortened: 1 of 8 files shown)


fideloper:
Good use of partials!

laravel/partials/_volume_storage_steps.html.md.erb
```diff
@@ -0,0 +1,40 @@
+1.  Make sure you are in your Laravel Fly App's directory
+
+    ```cmd
```
fideloper, line 3: Just double check the formatting when using both a tab (err 4 spaces or whatever) AND the back-tics for code blocks.

fideloper (approved):

laravel/partials/_volume_storage_steps.html.md.erb
KTanAug21: yes sir! checking this one

KTanAug21: update: 
I think the tabs help keep the numbering in the list,

 with tabs:
image

___________________________________________________________________________
but removing the tabs/spaces restart the numbering of the items: to 1,1, 
image

so probably tabs group sections under an item

PR #541, opened 2023-01-18
Fix incorrect tab on yaml file.

adamaiken89:
The sample file for GitHub actions contain some incorrect tabbing. This PR is to standardise all as 2 space for every tab.


jeregrine (approved):
Thank you so much! <3

PR #547, opened 2023-01-21
removes old behavior

byellokore:
I have ran in issue while trying to setup Sidekiq. Following the current version of the tutorial,  I got 3 `processes` after deploy.

* app
* web
* worker 

even without have any reference to app on my fly.toml. So, I believe this tutorial is not accurate anymore. Sorry if I'm wrong but I ended up with 3 'processes'. 

At least in the application information: I could even scale this third machine,
Screen Shot 2023-01-18 at 09 53 10

However in the monitoring window I had only two: web and worker. 

Screen Shot 2023-01-18 at 09 53 24

Anyhow, if we go with app and worker it's working fine.

Please close this if I'm wrong. 

btw - app is deleted now

### DIFF
--- a/rails/the-basics/sidekiq.html.md
+++ b/rails/the-basics/sidekiq.html.md
@@ -36,20 +36,10 @@ Add the following to the `fly.toml`:
 
 ```toml
 [processes]
-web = "bin/rails fly:server"
+app = "bin/rails fly:server"
 worker = "bundle exec sidekiq"
 ```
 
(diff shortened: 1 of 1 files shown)


jsierles:
@rubys What do you think about this?

bradgessler:
I cargo-culted the `web` process config from some other docs because I assumed `app` had a default behavior.

I actually prefer this approach if `app` can always remain and `worker` just an addition to the configuration. It reduces the number of steps people have to make to add workers to their apps.

👏👏👏

PR #555, opened 2023-01-24
Update sign-in.html.md.erb

nuno84:
This was done for windows. fly command doesnt exist.

### DIFF
--- a/hands-on/sign-in.html.md.erb
+++ b/hands-on/sign-in.html.md.erb
@@ -9,7 +9,7 @@ toc: false
 If you already have a Fly.io account, all you need to do is sign in with flyctl. Simply run:
 
 ```cmd
-fly auth login
+flyctl auth login
 ```
 
 Your browser will open up with the Fly.io sign-in screen, enter your user name and password to sign in. If you signed up with GitHub, use the `Sign in with GitHub` button to sign in.


catflydotio:
@nuno84 and @butlersrepos: The `fly` command should exist as a symlink, if you used brew or one of our install scripts to install flyctl. 

How did you each install flyctl? And on which OS?

catflydotio:
Thanks, @nuno84. I'll close this for now. Feel free to open a different issue if the installation doc has an error!

ndarilek:
Windows user and flyctl contributor here:

FWIW, I'm on Windows 10 and appear to have a fly symlink. If you'd not mind, I'd be interested in seeing a transcript of running the powershell installation script. Text-only please, no screenshots.

I *believe* admin permissions are required to make a symlink under Windows, but might be wrong. Could be these weren't granted, or that particular line failed. I'd like to fix it if at all possible, though.

Thanks!

nuno84:
Hi @ndarilek you nailed it!
I ran the script with powershell as administrator and it did the symlink trick.
So, yes (and for future reference) I confirm, the powershell MUST be executed with admin privileges to make the symlink, otherwise it doesn't.
Thank you once again.

ndarilek:
Awesome, thanks for confirming. I'll look into adding that to the script.

nuno84:
I would suggest to add that note on the fly.io CLI installation page:
[fly install](https://fly.io/docs/hands-on/install-flyctl/)
I was checking if I had missed that tip, but it wasn't mentioned there.

ndarilek:
The installation script now requests admin privileges to make the symlink if it doesn't exist. Guessing that should deploy soon with the next release.

Thanks for reporting!

PR #558, opened 2023-01-26
Update titles to keep consistency

katiayn:


catflydotio:
Background:  We used to have only one page per framework, just a quick launch example, called "Run a Whatever App". Our "happy path" frameworks (ones we have people working on--Elixir, Laravel, Rails, and now Django) sprouted their own whole sections. We should change main nav and the Laravel title to match Elixir and Rails (and add an .io to Elixir), though.

katiayn:
Update to:

- Elixir on [Fly.io](http://fly.io/) (Keep consistency with the [navbar](https://github.com/superfly/docs/blob/main/partials/_firecracker_nav.html.erb#L34))
- Rails on [Fly.io](http://fly.io/)
- Laravel on [Fly.io](http://fly.io/)
- Django on [Fly.io](http://fly.io/)

PR #593, opened 2023-02-16
Put soft_limit and hard_limit in code blocks

cwhite92:
Avoids markdown formatting italicizing text, and makes it consistent with the rest of the page


catflydotio:
Thanks @cwhite92

PR #597, opened 2023-02-18
Avoid confusion from "nginxproxy.fly.dev" not being a real address

benhoyt:
Currently this doc is confusing as it sounds like `nginxproxy.fly.dev` is a real domain name that your CNAME should point to ... but it's really talking about `<yourapp>.fly.dev`. I was confused by this and set up the wrong CNAME at first, and it looks like I'm [not the only one](https://community.fly.io/t/ssl-termination-error-with-custom-domain-cname-and-new-certificate/6754/4).


jackie-pc:
I hit this exactly.  Of course if I used my brain it would have been obvious.  But in a rush, I hit this issue!

andie787:
fixed by #1699

PR #599, opened 2023-02-21
Update static.html.md.erb

irishgeoff10:
added forms solution


catflydotio:
Thanks for the contribution, but this is a little too specific. 😄

PR #626, opened 2023-03-05
Update launch-app.html.md.erb

annamari12:
Remove link to the deleted page.


catflydotio:
Thanks @annamari12! This is fixed in an active PR that I don't want to generate conflicts with, so I'll close this.

PR #632, opened 2023-03-07
remove processes in mounts, and move possible errors under persisting…

KTanAug21:
… storage folder in laravel docs


KTanAug21:
This can be used in main as well, since the changes are not V2 specific. Will close this and open another request to merge with main instead!

PR #634, opened 2023-03-09
update fastify example to work with latest release

gregjopa:
I'm a big fan of Fly.io and recently deployed a Node.js app running Fastify on the platform. The latest release of Fastify (version 4) changed the signature of the listen function to `.listen([options][, callback])`. This PR updates the troubleshooting guide so the Fastify hello world example matches what's on the [Fastify home page](https://www.fastify.io/).

Fastify docs for listen(): https://github.com/fastify/fastify/blob/main/docs/Reference/Server.md#listen


jeregrine:
Awesome thank you!

PR #635, opened 2023-03-10
Modify steps to match CLI experience

allardbrain:
- The CLI now includes a question about Upstash Redis
- The CLI prompts for the first deployment, so running `flyctl deploy` after that in this example is redundant


allardbrain:
Hi there, I saw a couple of things that appear to be outdated in the docs while going through the CLI example. I went ahead and suggested changes; hopefully I got them correct! If you're interested in merging this but something is incorrect, just let me know and I'll change it. Thank you!

PR #636, opened 2023-03-13
Add documentation about the 60-second proxy timeout

jsierles:
The proxy timeout confuses people in [various](https://twitter.com/aweary/status/1633926337890078720) ways. So, here's a first shot at documenting it for public proxy services.


fideloper (commented):
:shipit:

PR #627, opened 2023-03-06
Apps v2/release

catflydotio:
Docs for when Apps V2 drops

Preview at https://appsv2-docpreview.fly.dev/docs/ (up to date at end of 24 March)

Most of the changes are in the addition of this section:
https://appsv2-docpreview.fly.dev/docs/apps/

PR #643, opened 2023-03-24
Mount section doesn't support processes parameter

guillaumervls:


catflydotio:
There's already a corresponding change in our [monster apps v2 PR](https://github.com/superfly/docs/pull/627), so I'll close this.

PR #645, opened 2023-03-27
rephrase multi services & processes section

guillaumervls:


catflydotio:
While we're in here, I'll suggest going further and grouping the information more concisely than I originally did; something like:

---

Chances are, you only want user requests to hit one of your processes; in the above example, that process is `web`, so you would specify the `web` process in the [`[[services]]` section](/docs/reference/configuration/#the-services-sections) for the app's HTTP service:

```
...
[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["web"]
  ...
```

You can define distinct services for each process that needs to accept connections via Fly Proxy (whether [publicly](/docs/reference/services/) or via [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing)), by creating multiple `[[services]]` sections in `fly.toml`. 

**Note:** Make sure processes handle connections on different [external ports](/docs/reference/configuration/#services-ports). Fly Proxy doesn't know about process groups; it load-balances requests among all Machines with a service configured on the requested port.

---

✅ Emphasizes what the gotcha might be before explaining the behaviour of Fly Proxy
✅  Links to the doc where you can learn more about setting external ports. 

What do you think? Does this address your concerns?

PR #653, opened 2023-04-03
Jphenow/shared ipv4

jphenow:
I'm not confident that explanation of shared IPs won't unnecessarily freak someone out

### DIFF
--- a/about/pricing.html.md.erb
+++ b/about/pricing.html.md.erb
@@ -21,7 +21,7 @@ Our free allowance applies to all plans, including the Hobby plan, and includes 
 
 If you want to scale beyond the included free resources, you can pay for just what you need at the usage-based pricing listed below.
 
-If you need more support or compliance options, you can choose one of our [paid plans](https://fly.io/plans). These come with usage included and additional support options. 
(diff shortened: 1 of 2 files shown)


about/pricing.html.md.erb
```diff
@@ -64,7 +64,9 @@ Volume billing is pro-rated to the hour.
 
 ### Anycast IP addresses
 
-Each application includes one Anycast IPv4 and unlimited Anycast IPv6 Anycast addresses for global load balancing. Additional IPv4 addresses are $2/mo.
+Each application receives a [shared IPv4 address](/docs/reference/services/#shared-ipv4) and unlimited [Anycast IPv6](/docs/reference/services/#ipv6) addresses for global load balancing.
+
+Dedicated IPv4 addresses are $2/mo.
```
bzanchet, line 69: Not sure if we know the answer, but I'd like to see something akin to `Volume billing is pro-rated to the hour.` here, just to be clear what's being charged there.

jphenow: Do we know that we're going to pro-rate them?

bzanchet: ah ok-- if they're not, I'd rather read `ips are not pro-rated` etc..

PR #654, opened 2023-04-03
WebP-a-GO Docs edition

jeregrine:


catflydotio:
I'm a bit lost. Are there webp images ready for these links to hit?

PR #664, opened 2023-04-10
Add missing word

beeburrt:
In first sentence - `about/pricing.html.md.erb`

I wasn't 100% sure on which word to use. "build" ? or "deploy" might be better. I went with "launch". Hope that's ok :)


catflydotio:
Thanks @beeburrt

PR #665, opened 2023-04-11
Retool over SSH Tunnel to postgres

jphenow:
I'm not positive we care a ton about _Retool_ but this the bones of connecting over SSH tunnel to a number of BI tools I think.

This is kind of choppy in its format it feels like. Also describing the steps for the Retool UI was painful - need some visual cues. 

Requesting review mostly for early feedback - your call if you invest more than 2m

### DIFF
--- a/app-guides/connect-retool.html.md
+++ b/app-guides/connect-retool.html.md
@@ -1,0 +1,123 @@
+---
+title: "Connect Retool to your Fly.io Database"
+layout: docs
+sitemap: false
+nav: firecracker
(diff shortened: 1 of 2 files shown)


jphenow:
wrote a fresh produce instead because this would fast and poorly:

* https://flyio.discourse.team/t/guides-and-other-docs-susceptible-to-rot/2485/6
* https://community.fly.io/t/connect-retool-to-your-fly-io-database/12470/1

PR #678, opened 2023-04-17
Fixing example volume creation region

bzanchet:
Fixing text referring to the example on the docs.

As pointed out here: https://community.fly.io/t/documentation-heavily-outdated-for-volumes/12288/3

### DIFF
--- a/reference/volumes.html.md.erb
+++ b/reference/volumes.html.md.erb
@@ -43,7 +43,7 @@ Every Fly Volume belongs to a [Fly App](/docs/reference/apps/).
 
 Create a volume for an app using `fly volumes create`. The default volume size is 3GB. See [`fly volumes create`](/docs/flyctl/volumes-create/) in the [flyctl reference](/docs/flyctl) for usage and options.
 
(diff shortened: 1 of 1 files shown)


reference/volumes.html.md.erb
```diff
@@ -61,7 +61,7 @@ Created at: 04 Mar 23 03:32 UTC
 
 Volumes are, by default, created with encryption-at-rest enabled for additional protection of the data on the volume. Use `--no-encryption` to instead create an unencrypted volume for improved performance at deployment and runtime.
 
-Volumes are bound to both apps and regions. A volume is directly associated with only one app and exists in only one region. No other app can see this volume and only an instance of the app running in the LHR region can access it.
+Volumes are bound to both apps and regions. A volume is directly associated with only one app and exists in only one region. No other app can see this volume and only an instance of the app running in the YYZ region can access it.
```
catflydotio, line 64:
```suggestion
Volumes are bound to both apps and regions. A volume is directly associated with only one app and exists in only one region. No other app can see this volume and only an instance of the app running in the yyz region can access it.
```
for consistency in this page

catflydotio (approved):

PR #688, opened 2023-04-19
Add autostart and stop docs

andie787:
This PR: 
- adds a new page under Apps about automatically stopping and starting Machines
- adds the new settings to the fly.toml config doc page
- adds new page to sidenav and apps index page

Source material: https://community.fly.io/t/automatically-starting-stopping-apps-v2-instances/12342

Tested locally. See comments for questions/issues.

### DIFF
--- a/apps/autostart-stop.html.md.erb
+++ b/apps/autostart-stop.html.md.erb
@@ -1,5 +1,5 @@
 ---
-title: Automatically Stop and Start App Machines
+title: Automatically Stop and Start App V2 Machines
 objective: 
 layout: docs
 nav: firecracker
@@ -13,8 +13,8 @@ You can decide whether you want to automatically stop, or downscale, instances o
 ```toml
 ...
 [[services]]
-  autostart = true
-  autostop = true
+  auto_stop_machines = true
+  auto_start_machines = true
 ...
 ```
 
@@ -28,16 +28,16 @@ You can decide whether you want to automatically stop, or downscale, instances o
 ...
 ```
 
-## Default and recomended values
+## Default and recommended values
 
-New V2 apps are automatically started and automatically stopped by default:
+New V2 app created using the `fly launch` command are automatically started and automatically stopped by default:
 
 ```toml
 auto_stop_machines = true
 auto_start_machines = true
 ```
 
-Existing apps are automatically started but not automatically stopped by default:
+Existing apps—or any apps that don't have these settings in `fly.toml`—are automatically started but not automatically stopped by default:
 
 ```toml
 auto_stop_machines = false
--- a/apps/autostart-stop.html.md.erb
+++ b/apps/autostart-stop.html.md.erb
@@ -10,11 +10,15 @@ order: 65
 This feature only works for V2 apps. For information about scaling V1 apps, refer to [Scale V1 Nomad Apps](/docs/apps/legacy-scaling/).
 </div>
 
-You can decide whether you want to automatically stop an app's Fly Machines when they're idle and automatically restart them when they're needed. Fly Proxy controls automatically starting and automatically stopping Machines according to settings in your `fly.toml` file, in either the `services` or `http_service` sections:
+Fly Machines are fast to start and stop, and you don't pay for their CPU and RAM when they're in the `stopped` state. For Fly Apps with a service configured, Fly Proxy can start and stop existing Machines based on incoming requests, so that your app can accommodate bursts in demand without keeping extra Machines running constantly.
+
+You can configure automatic starts and automatic stops independently with the `auto_stop_machines` and `auto_start_machines` settings, and tune them with concurrency limits, within the [[[services]]](/docs/reference/configuration/#the-services-sections) or [[http_service]](/docs/reference/configuration/#the-http_service-section) sections of `fly.toml`.
 
 ```toml
 ...
 [[services]]
+  internal_port = 8080
+  protocol = "tcp"
   auto_stop_machines = true
   auto_start_machines = true
 ...
@@ -56,7 +60,7 @@ If `auto_start_machines = false` and `auto_stop_machines = true`, Fly Proxy will
 
 The Fly Proxy runs a process to automatically stop and start Fly Machines every few minutes.
 
-When `auto_stop_machines = true` in your `fly.toml`, the proxy looks at Machines running in a single region and uses the [`soft_limit` setting](/docs/reference/configuration/#the-http_service-section) for each Machine to determine if there's excess capacity. If the proxy decides there's excess capcity, it will stop exactly one machine. The proxy repeats this process every few minutes, stopping only one machine per region, if needed, each time.
+When `auto_stop_machines = true` in your `fly.toml`, the proxy looks at Machines running in a single region and uses the [`soft_limit` setting](/docs/reference/configuration/#the-http_service-section) for each Machine to determine if there's excess capacity. If the proxy decides there's excess capacity, it will stop exactly one machine. The proxy repeats this process every few minutes, stopping only one machine per region, if needed, each time.
 
 Fly Proxy considers all of the app's Machines that are currently running in a given region, such as `fra`, and determines excess capacity as follows:
 
@@ -67,7 +71,7 @@ Fly Proxy considers all of the app's Machines that are currently running in a gi
   * the proxy checks if the Machine has any traffic
   * if the Machine has no traffic (a load of 0), then the proxy stops the Machine
 
-When `auto_start_machines = true` in your `fly.toml`, the Fly Proxy restarts an a Machine in the nearest region when required.
+When `auto_start_machines = true` in your `fly.toml`, the Fly Proxy restarts a Machine in the nearest region when required.
 
 Fly Proxy determines when to start a Machine as follows:
 
--- a/apps/autostart-stop.html.md.erb
+++ b/apps/autostart-stop.html.md.erb
@@ -1,5 +1,5 @@
 ---
-title: Automatically Stop and Start App V2 Machines
+title: Automatically Stop and Start App V2 Fly Machines
 objective: 
 layout: docs
 nav: firecracker
@@ -7,10 +7,10 @@ order: 65
 ---
 
 <div class="callout">
-This feature only works for V2 apps. For more information about scaling V1 apps, refer to [Scale V1 Nomad Apps](/docs/apps/legacy-scaling/).
+This feature only works for V2 apps. For information about scaling V1 apps, refer to [Scale V1 Nomad Apps](/docs/apps/legacy-scaling/).
 </div>
 
-You can decide whether you want to automatically stop an app's machines when they're idle and automatically restart them when they're needed. Fly Proxy controls automatically starting and automatically stopping machines according to settings in your `fly.toml` file, in either the `services` or `http_service` sections:
+You can decide whether you want to automatically stop an app's Fly Machines when they're idle and automatically restart them when they're needed. Fly Proxy controls automatically starting and automatically stopping Machines according to settings in your `fly.toml` file, in either the `services` or `http_service` sections:
 
 ```toml
 ...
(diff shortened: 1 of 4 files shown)


reference/configuration.html.md.erb
```diff
@@ -206,6 +208,12 @@ As with the more verbose [`[[services]]`](#the-services-sections), you can speci
...
 * `force_https`: A boolean which determines whether to enforce HTTP to HTTPS redirects.
+* `auto_stop_machines`: Whether to automatically stop instances of an application when there's excess capacity, per region. If there's only one instance in a region, then the instance is stopped if it has no traffic when the autostop process runs. The default it `true`.
+* `auto_start_machines`: Whether to automatically start instances of an application when a new request is made to the application and there's no excess capacity, per region. If there's only one instance in a region, then it's started whenever a request is made to the application. The default is `true`.
+
+<div class="callout">
+We recommend setting `auto_stop_machines` and `auto_start_machines` to the same value to avoid having instances that either never start or never stop. Learn more about [automatically starting and stopping Fly machines](/docs/apps/autostart-stop/), including how the process determines excess capacity in a region using the `soft_limit` setting.
+</div>
```
andie787, line 216: Is this ok as a callout?

andie787, line 216: Locally, the link within the callout was invisible (no underline). Is that normal?

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
+---
+title: Automatically Stop and Start App Machines
```
dangra, line 2: Should it be "App V2" ? somehow we have to make it clear this is a V2 only feature.

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+
+<%= partial "/docs/partials/v2_transition_banner" %>
+
+You can decide whether you want to automatically stop, or downscale, instances of your app when they're idle and automatically restart them when they're needed. Each instance of your app is running on a Fly machine. Fly Proxy controls automatically starting and automaticaly stopping machines according to settings in your `fly.toml` file, in either the `services` or `http_service` sections:
+
+```toml
+...
+[[services]]
+  autostart = true
+  autostop = true
```
dangra, line 17: this should be `auto_start_machines` and `auto_stop_machines`

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+## Default and recomended values
+
+New V2 apps are automatically started and automatically stopped by default:
+
+```toml
+auto_stop_machines = true
+auto_start_machines = true
+```
+
+Existing apps are automatically started but not automatically stopped by default:
```
dangra, line 40: Truly speaking, Any service without `auto_start(stop)_machines` fields set in fly.toml will use these defaults.
That is the case for existent apps because they don't have them.

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+  internal_port = 8080
+  force_https = true
+  auto_stop_machines = true
+  auto_start_machines = true
+...
+```
+
+## Default and recomended values
+
+New V2 apps are automatically started and automatically stopped by default:
```
dangra, line 33: Actually, new V2 apps created by `fly launch` since flyctl v0.0.527 will get these fields for its generated fly.toml

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+Existing apps are automatically started but not automatically stopped by default:
+
+```toml
+auto_stop_machines = false
+auto_start_machines = true
+```
+
+In most cases, we recommend setting `auto_stop_machines` and `auto_start_machines` to the same value to avoid having instances that either never start or never stop. 
+
+If `auto_start_machines = true` and `auto_stop_machines = false`, the proxy will start your instances but they will never be stopped automatically. If you have a reason to only stop instances manually, then these settings are just fine.
```
dangra, line 49: I think th' is is valid case and we have a chance to link to a form post or even better a doc explaining how to use "self stopping" machines.

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+
+Fly proxy determines when to start an instance as follows:
+
+* The proxy waits for a request to your app.
+* If all running instances are above their `soft_limit` setting, then the proxy starts a stopped instance from the nearest region (if there are any stopped instances).
+* The proxy routes the request to the newly started instance.
+
+## When to Stop and Start Machines Automatically, or Not
+
+If your app has highly variable workloads, then you can set `auto_stop_machines` and `auto_start_machines` to true to manage your instances as demand decreases and increases. This could reduce costs, because you'll never have to run excess instances to handle peak load; you'll only run what's necessary at any given time.
```
dangra, line 78: should we mention to make it clear that cost reduction is due to not charging for "stopped" machines?

reference/configuration.html.md.erb
dangra: looks good to me. The link to the doc that explain it further is even better.

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+title: Automatically Stop and Start App Machines
+objective: 
+layout: docs
+nav: firecracker
+order: 65
+---
+
+<%= partial "/docs/partials/v2_transition_banner" %>
+
+You can decide whether you want to automatically stop, or downscale, instances of your app when they're idle and automatically restart them when they're needed. Each instance of your app is running on a Fly machine. Fly Proxy controls automatically starting and automaticaly stopping machines according to settings in your `fly.toml` file, in either the `services` or `http_service` sections:
```
catflydotio, line 11: We need to mention that this applies only to V2 apps.  Autoscaling in Apps V1 is Nomad-controlled. Also doesn't really work but see https://fly.io/docs/apps/legacy-scaling/ for scaling of V1 apps.

(also whoops https://fly.io/docs/flyctl/autoscale/ we need to tell flyctl to say that's only for V1 apps) 

It'll be confusing to call this downscaling. It stops a running machine but leaves it in place. We're not calling this autoscaling in general. It only works within the set of machines you’ve already provisioned. `fly scale count` (which now exists for V2: https://github.com/superfly/flyctl/pull/1983) creates or destroys Machines.

Note that we don't call Machines instances. It's too Nomady.

(typo automaticaly)

reference/configuration.html.md.erb
catflydotio: We need to ask Kyle for a style for links in a callout.

apps/autostart-stop.html.md.erb
catflydotio: Another point: this is a feature of Fly Proxy, so only works on Machines with services configured. (You can still get a Machine with or without services to stop by having its main process exit)

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+
+If `auto_start_machines = false` and `auto_stop_machines = true`, the proxy will scale your instances down but won't be able to start them again. If all of your instances are scaled down, then requests will start failing.
+
+## How It Works
+
+The Fly Proxy runs the process to stop and start machines automatically every few minutes.
+
+When `auto_stop_machines = true` in your `fly.toml`, the proxy looks at machines running in a single region and uses the [`soft_limit` setting](/docs/reference/configuration/#the-http_service-section) for each machine to determine if there's excess capacity. If the proxy decides there's excess capcity, it will stop exactly one machine. The proxy repeats this process every few minutes, stopping only one machine per region, if needed, each time.
+
+Fly proxy considers all of the running instances of your app in a given region, such as `fra`, and determines excess capacity as follows:
```
catflydotio, line 59: Fly proxy -> Fly Proxy, unless we want to call it fly-proxy like its repository — or something else — a branding decision. I’ve been calling it Fly Proxy.

apps/autostart-stop.html.md.erb
catflydotio, line 78: It's all about connections or requests to a configured service here (no CPU or RAM load taken into account here) -- so we should make that clear -- it's not a general workload.

true -> `true`; false-> `false` makes it easier to parse

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,83 @@
...
+...
+[http_service]
+  internal_port = 8080
+  force_https = true
+  auto_stop_machines = true
+  auto_start_machines = true
+...
+```
+
+## Default and recomended values
```
catflydotio, line 31:
```suggestion
## Default and recommended values
```

apps/autostart-stop.html.md.erb
andie787: Should we just say "created by `fly launch`" then, or do we usually include versions like that?

apps/autostart-stop.html.md.erb
andie787: @dangra Do you have something in mind or is it not written yet? 🙂

apps/autostart-stop.html.md.erb
andie787: @catflydotio 
So, do we remove the banner then as well?

I got "downscaling" from some logs in this community post: https://community.fly.io/t/what-does-downscaling-app-mean/12309. I wanted to connect the terms somehow. Thoughts?

apps/autostart-stop.html.md.erb
dangra: I will leave @catflydotio to answer that 🐱

apps/autostart-stop.html.md.erb
dangra: @senyosimpson gave a good reason at https://community.fly.io/t/automatically-starting-stopping-apps-v2-instances/12342/5?u=dangra

apps/autostart-stop.html.md.erb
catflydotio: > So, do we remove the banner then as well?

Since nothing in this doc applies to Nomad apps, we could have a banner or callout saying "this applies to V2 apps only; see xyz doc for scaling V1 apps" instead of the "rolling out v2 banner" (which itself could use an update)

> Are you saying replace all instances of "instance" with "machine" in this doc?

Where the sentence still makes sense, yes.

apps/autostart-stop.html.md.erb
catflydotio: I was mulling that. I don't have a perfect answer. "Our proxy" can also be handy? Since there are a few things "the proxy" could mean depending on the context, we just have to keep an eye out for any ambiguity.

If it's clear from the start that this whole feature is a feature of Fly Proxy, then it hopefully sets people up to think about our proxy first and not some other kind.

apps/autostart-stop.html.md.erb
catflydotio: > I got "downscaling" from some logs in this community post: https://community.fly.io/t/what-does-downscaling-app-mean/12309. I wanted to connect the terms somehow. Thoughts?

My thought on this is that the logs shouldn't call it downscaling. :)

apps/autostart-stop.html.md.erb
catflydotio: I think there's more to say here.

If this part of `fly launch` hasn't changed recently: 

Launching creates a config, and writes it into `fly.toml`; the Machines in the app get configured according to that config if you let `fly launch` carry on through deployment.

But if you say `N` to "deploy now?" the Machines are configured according to what's in `fly.toml` at the time.
You can edit `fly.toml` before the first deployment, and the app will have whatever settings you specified.

Incidentally, this has been kind of a gotcha: `fly launch` writes you a `fly.toml` and is waiting for you to tell it whether to deploy or not, and you can edit and save `fly.toml`, but it's got a config in memory already and doesn't read `fly.toml` again before it deploys.

All that to say: what `fly launch` puts into a new app's config is one kind of default.

You can create an app without `fly launch` (`fly apps create` and write your own `fly.toml`) and it won't be configured that way automatically.

Once people understand that, it's not too serious which version they're using (these settings have only been generated at all for like 5 minutes, anyway, right?). I don't object to putting version numbers in, if you think it's important.

All that to say: what about 

```
`fly launch` configures new V2 apps to start and stop automatically:
```

(Or put it passively to avoid putting `fly launch` at the start of the sentence if you like; I find it useful to just let commands go at the starts of sentences)

apps/autostart-stop.html.md.erb
```diff
@@ -0,0 +1,84 @@
...
+layout: docs
+nav: firecracker
+order: 65
+---
+
+<div class="callout">
+This feature only works for V2 apps. For information about scaling V1 apps, refer to [Scale V1 Nomad Apps](/docs/apps/legacy-scaling/).
+</div>
+
+You can decide whether you want to automatically stop an app's Fly Machines when they're idle and automatically restart them when they're needed. Fly Proxy controls automatically starting and automatically stopping Machines according to settings in your `fly.toml` file, in either the `services` or `http_service` sections:
```
catflydotio, line 13: Suggested fodder:

Fly Machines are fast to start and stop, and you don't pay for their CPU and RAM when they're in the `stopped` state. For Fly Apps with a [service](/docs/reference/configuration/#the-services-sections) configured, Fly Proxy can start and stop existing Machines based on incoming requests, so that your app can accommodate bursts in demand without keeping extra Machines running constantly.

You can configure automatic starts and automatic stops independently with the `auto_start_machines` and `auto_stop_machines` settings within the `[[services]]` or `[http_service]` sections of `fly.toml`, and tune them with the service's [concurrency limits](/docs/reference/configuration/#services-concurrency).

partials/_firecracker_nav.html.erb
```diff
@@ -64,6 +64,8 @@
       <li>
         <%= nav_link "Scale the Number of Machines", "/docs/apps/scale-count/" %>
       </li>
+      <li>
+        <% nav_link "Auto Stop and Start Machines", "/docs/apps/autostart-stop/" %>
```
catflydotio, line 68:
```suggestion
        <% nav_link "Auto Stop and Start Machines", "/docs/apps/autostart-stop/" %>
    </li>
```

catflydotio (commented):
Made a suggestion on the introduction and added a missing tag. Anything else could be tweaked later.

apps/autostart-stop.html.md.erb
andie787: This is great. Tweaked a bit to get links lined up.

PR #690, opened 2023-04-21
mention that ENTRYPOINT always runs after release_command

brenol:
Hello!

I had an issue today with my deployments not finishing  and I ended up opening an issue as I had no idea what was causing it:
https://community.fly.io/t/deployment-issue-release-command-timing-out/12395/7

Here's the PR that adds to the release_command documentation that it only overrides `CMD`, and that `ENTRYPOINT` will always run.

Wasn't sure where to put it, so I left it right at the start of the docs. But I also think it could be near the `RELEASE_COMMAND` env var, so I just went with what had the smallest change overall.


brenol:
Closing as #691 expands the release_command documentation.

PR #691, opened 2023-04-21
config doc: clarify how release_command works

catflydotio:
@brenol pointed out some clarifications for the release command option in fly.toml. (https://github.com/superfly/docs/pull/690) 

Github's not letting me make suggestions on the PR and I wanted to change some more things, thus this new PR

### DIFF
--- a/reference/configuration.html.md.erb
+++ b/reference/configuration.html.md.erb
@@ -138,7 +138,7 @@ A non-zero exit status from this command will stop the deployment. `fly deploy` 
 
 To ensure the command runs in a specific region - say `dfw` - set `PRIMARY_REGION = 'dfw'` on in your application environment in `fly.toml` or with `fly deploy -e PRIMARY_REGION=dfw`. Setting `PRIMARY_REGION` is important if when running [database replicas in multiple regions](/docs/postgres/high-availability-and-global-replication).
 
-The environment variable `RELEASE_COMMAND=1` is set for you within the temporary release VM. This might be useful if you need to customize your Dockerfile `ENTRYPOINT` to behave differently within a release VM.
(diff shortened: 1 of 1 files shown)


brenol:
Thanks! I knew it would be way better with more information from you guys, so thank you. :) I'll close my PR.

reference/configuration.html.md.erb
```diff
@@ -124,20 +124,21 @@ This section configures deployment-related settings such as the release command
...
 ```
+The `release_command` value replaces `CMD` in the temporary VM. This is useful for running database migrations before app VMs are created or updated with the new release. Note that the Docker image's `ENTRYPOINT` is not overridden by the `release_command`; `ENTRYPOINT` always runs.
 
-This command runs in a temporary VM - using the successfully built release - *before* that release is deployed. This is useful for running database migrations.
-
-The temporary VM has full access to the network, environment variables and secrets, but *not* to persistent volumes.  Changes made to the filesystem on the temporary VM will not be retained or deployed.  If you need to modify persistent volumes or configure your application, consider making use of `CMD` or `ENTRYPOINT` in your Dockerfile.
+The temporary VM has full access to the network, environment variables and secrets, but *not* to persistent volumes. Changes made to the filesystem on the temporary VM will not be retained or deployed.  If you need to modify persistent volumes or configure your application, consider making use of `CMD` in your Dockerfile, or of a [process group command](#the-processes-section).
```
catflydotio, line 135:
```suggestion
The temporary VM has full access to the network, environment variables and secrets, but *not* to persistent volumes. Changes made to the filesystem on the temporary VM will not be retained or deployed.  If you need to modify persistent volumes or configure your application, consider making use of `CMD` or `ENTRYPOINT` in your Dockerfile, or of a [process group command](#the-processes-section).
```

PR #704, opened 2023-04-26
Update volume docs for clarity and user feedback

andie787:
**Work in Progress**
- The goal of the structure changes is to move content around so that the "Databases & Storage" page is an overview of all storage options, "Volumes reference" is where you find all the details, and "Add Volume Storage" comes after you've chosen storage and read the Volume reference (hopefully).
- Consolidate and make descriptions consistent
- These community posts gave us some direction on what was needed: https://community.fly.io/t/documentation-heavily-outdated-for-volumes/12288 and https://community.fly.io/t/some-issues-with-volumes-on-apps-v2/11994

### [Databases and Storage](https://fly.io/docs/database-storage-guides/) page
Preview: https://andrea-docs-preview.fly.dev/docs/database-storage-guides/
- moved section on `choosing your approach to storage` here
- minor edits

### [Volumes](https://fly.io/docs/reference/volumes/) reference page
Preview: https://andrea-docs-preview.fly.dev/docs/reference/volumes/
- remove bolded "this is a reference doc", to be replaced by other links. Machines section doesn't have volume info in it in any case. BUt maybe we need extra links back the app section?
- add exceptions to the rule of volumes per [https://community.fly.io/t/multiple-fly-volumes-confusion/12329](https://community.fly.io/t/multiple-fly-volumes-confusion/12329) and [https://community.fly.io/t/so-its-not-possible-to-use-sqlite-on-fly-without-litefs/12306](https://community.fly.io/t/so-its-not-possible-to-use-sqlite-on-fly-without-litefs/12306)
- make each flyctl command section consistent, all sections edited/updated
- changed delete a volume to destroy a volume
- UPDATE: added sections for Access a Volume and Mount a Volume
- UPDATE: put the reference content under another heading: Working with Volumes and a link to Add and Use Volumes for the step-by-step. This should avoid the scenario where users try to "follow" this reference.

### [Add Volume Storage](https://fly.io/docs/apps/volume-storage/) (changed to Use Volume Storage) page
Preview: https://andrea-docs-preview.fly.dev/docs/apps/volume-storage/
- added status command output to the `Deploy the app` step

TBD in new PR:
- added section on using cloning to add a volume to an existing app (still testing, since it doesn't work yet without this PR: https://github.com/superfly/flyctl/pull/2190 and the text needs to be augmented with more details and outputs)

### DIFF
--- a/apps/volume-storage.html.md.erb
+++ b/apps/volume-storage.html.md.erb
@@ -10,11 +10,11 @@ order: 30
 
 Volumes are local persistent storage for [Fly Machines](/docs/machines/). Volumes allow an app to save its state—to preserve configuration, session or user data—and then be restarted with that information in place.
 
-**The first rule of Fly Volumes is: always run at least two of them per application.** The way to do this is to run at least two Machines per app. Also note that volumes don't sync up by themselves. Your app need to take care of that. Refer to [Fly Volumes](/docs/reference/volumes/) for exceptions to the rule and details about how volumes are implemented.
+**The first rule of Fly Volumes is: always run at least two of them per application.** The way to do this is to run at least two Machines per app. Also note that volumes don't sync up by themselves. Your app needs to take care of that. Refer to [Fly Volumes](/docs/reference/volumes/) for exceptions to the rule and details about how volumes are implemented.
 
 ## Launch an app with a Fly Volume
 
-Follow these steps to create a new app, a new volume, and then mount the volume before deply. If you need to add a volume to an exsiting app, refer to TBD. 
+Follow these steps to create a new app, a new volume, and then mount the volume before deploy. If you need to add a volume to an existing app, refer to [Add a volume to an existing app by cloning](#add-a-volume-to-an-existing-app-by-cloning). 
 
 ### Launch the app, but don't deploy it yet
 
@@ -130,7 +130,32 @@ Create an additional volume in the desired region, then [use `fly scale count`](
 
 ## Add a volume to an existing app by cloning
 
-TBD since there are 2 open PRs to possibly improve this process
+If you try to add a volume to an app that's already been deployed, you might get errors because the volume and the Machine created for your app on deploy need to be on the same physical host. One way to get around this is to attach a new volume to a clone of one of your app's Machines.
+
+Create new volume for your app in the same region the app will be deployed:
+
+```cmd
+fly volumes create myapp_data --region lhr --size 1 --app myapp
+```
+
+Clone one of your app's Machine (with no volume) and attach the volume you just created:
+
+```cmd
+fly machine clone <machine-id> --attach-volume <vol-id>
+```
+
+Destroy the Machine you used to create the clone:
+```cmd
+fly machine destroy <non-volume-machine-id>
+```
+
+Check whether the volume is properly attached to the Machine:
+
+```cmd
+fly volumes list
+```
+
+The volume you just created should have the newly-cloned Machine's ID listed under `ATTACHED VM`.
 
 ## Remove a volume from an app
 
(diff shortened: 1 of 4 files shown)


reference/volumes.html.md.erb
```diff
@@ -106,15 +102,17 @@ Created at: 04 Mar 23 03:32 UTC
...
 ```cmd
-fly volumes extend <volume-id> -s <new-size>
+fly volumes extend vol_od56vjp5pzmvny30 -s 2
 ```
 
-where `<new-size>` is the desired size in GB. 
+where `<new-size>` is the extended size in GB. 
 
-The Machine attached to the target volume must be restarted to allow the file system to be resized. **On legacy Nomad-orchestrated Apps, the instance is automatically restarted, so this step isn't needed.**
+You'll need to restart the Machine attached to the target volume to allow the file system to be resized. On Nomad (V1) Apps, the instance is automatically restarted, so this step isn't needed.
```
catflydotio, line 115:
```suggestion
You'll need to restart the Machine attached to the target volume to allow the file system to be resized. On V1 (Nomad) apps, the instance is automatically restarted, so this step isn't needed.
```

I'm not convinced that this last sentence doesn't need to be bold, or have a callout.

apps/volume-storage.html.md.erb
```diff
@@ -43,34 +27,52 @@ fly launch
...
 ```
 
+### Provision the volume
+
+Create a volume for the app, with the name you chose, in the same region you're deploying the app to. For example, the following command creates a volume called `myapp_data`, in region `lhr`, with 1GB of storage, for the app named `myapp`:
+
+```cmd
+fly volumes create myapp_data --region lhr --size 1 --app myapp
+```
+
```
catflydotio, line 41: We don't have to do this part anymore, on initial launch, do we? It gets hairier when you want to add a volume where there was already a Machine, but that's the in-flux part you were talking about

reference/volumes.html.md.erb
andie787: @catflydotio I can re-bold it. However, a note might be a better pattern for this kind of thing. Style experts discourage using bold for this scenario. :)

apps/volume-storage.html.md.erb
andie787: When or where do you create the volume to attach then?

catflydotio: At first deployment, if there's a mounts section in fly.toml, I believe a new volume is created for the new machine that's also created. Once there are machines on the app, then you have to clone a machine that has a volume, or restart with no machines--but making this smoother is what @dangra is taking a stab at with https://github.com/superfly/flyctl/pull/2189, I believe.

dangra: That is right, `fly deploy` will create a volume per process group that has a mount section defined in fly.toml **only on first deploy**.

And #2189 is about supporting the case of a user adding mounts after the first deploy. It won't create the volumes automatically, it will complain saying how many volumes are missing and ask the user to run `fly vol create` before attempting to deploy again.

andie787: Testing this now.

andie787: @dangra So is https://github.com/superfly/flyctl/pull/2189 the solution to adding a volume to an existing machine? You will need to create the volumes you need, update the fly.toml mounts section, and then run `fly deploy`? If you have, say, 4 machines in one process, will it add volumes to all of them (as long as 4 volumes are already created)?

dangra: Yes, that is right.

dangra: And I just merged it so you can update docs right away.

andie787: I'm going to merge this PR and open a new one for the changes in https://github.com/superfly/flyctl/pull/2189.

PR #705, opened 2023-04-27
Simplify static site instructions

a-h:
The static site documentation gets you to edit the `fly.toml` file to update the port.

It's simpler to update the Dockerfile to get `goStatic` to use port 8080 instead.

This shortens the instructions by removing a step.


andie787:
This is simpler and that's a good thing. But for this doc, I think it's a good way to introduce new users to the `fly.toml` config file.

PR #709, opened 2023-04-27
Let readers know configuration can also be changed via fly.toml file …

KTanAug21:
…for fly machines


KTanAug21:
Not applicable for fly machines, only for fly apps!

PR #714, opened 2023-05-01
Update flyctl_ssh_console.md

vyedin:
It's not super clear how to use this the first time to connect to an app, so I added a note to add the credential for the next person.


catflydotio:
These docs have to be updated from their source in the flyctl repo or the next flyctl update will overwrite them.

vyedin:
@catflydotio oh interesting. Do we want to take away the github link at the bottom then?

catflydotio:
@vyedin Right now flyctl command docs use the docs template -- it's fine for all other docs and rarely does anyone try to edit one. So I'm not opposed but not sure it's a big deal either.

catflydotio:
Actually it might be really easy to change on flyctl docs. Let me have a look

catflydotio:
How's it look now? (https://github.com/superfly/docs/commit/a34e0fd98b07921de28cc65debf51f561d1bdf16)

PR #716, opened 2023-05-02
Update volume docs with more ways to add volumes

andie787:
Add 2 ways to add volumes to existing apps (apps that have already been deployed). 

- clone and use --attach-volume, specifying the mount point
- set up volumes and run `fly deploy`

Other minor edits. Like renaming Add and Use Volume Storage back to just Add Volume Storage. We need to add more info about usage or maybe link to the specific framework guides that talk about volumes?

### DIFF
--- a/apps/volume-storage.html.md.erb
+++ b/apps/volume-storage.html.md.erb
@@ -118,7 +118,7 @@ At this point there are two identically configured Machines on the app, each wit
 
 ### On a V1 (Nomad) app
 
-Create an additional volume in the desired region, then [use `fly scale count`](/docs/legacy-scaling/) to add a VM to use it.
+Create an additional volume in the desired region, then [use `fly scale count`](/docs/apps/legacy-scaling/) to add a VM to use it.
 
 ## Add volumes to an existing app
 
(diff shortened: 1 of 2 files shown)


catflydotio:
A note about a callout in the volumes reference doc that doesn't seem to be new in this PR: the primary reason we say to run at least two machines (and two volumes) is host and network failures. Downtime on deployment is a good point, but people really complain when there's some sort of failure and their app is down for many minutes or hours because they didn't build in any redundancy.

reference/volumes.html.md.erb
```diff
@@ -109,7 +107,7 @@ Created at: 04 Mar 23 03:32 UTC
 
 ### Mount a Volume
```
catflydotio, line 108: I'm not sure about including how to mount or access a volume in the volumes reference doc. `fly.toml` is app-platform-as-a-service (i.e. `fly deploy`) -specific. You can mount volumes to non-platform Machines (in apps that have no `fly.toml`) too. I think it's fine to point to other docs from here though.

apps/volume-storage.html.md.erb
```diff
@@ -8,13 +8,120 @@ order: 30
 
 <%= partial "/docs/partials/v2_transition_banner" %>
 
-Volumes are local persistent storage for [Fly Machines](/docs/machines/). Volumes allow an app to save its state—to preserve configuration, session or user data—and then be restarted with that information in place.
+Volumes are local persistent storage for [Fly Machines](/docs/machines/). Volumes allow an app to save its state&mdash;to preserve configuration, session or user data&mdash;and then be restarted with that information in place. You can access and write to a volume on a Machine just like a regular directory.
 
 **The first rule of Fly Volumes is: always run at least two of them per application.** The way to do this is to run at least two Machines per app. Also note that volumes don't sync up by themselves. Your app needs to take care of that. Refer to [Fly Volumes](/docs/reference/volumes/) for exceptions to the rule and details about how volumes are implemented.
 
-## Launch an app with a Fly Volume
+## Add volumes to an existing app
```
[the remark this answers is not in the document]
andie787: Debated about this as well. Decided to put "add" first since that's what we hear about in Community posts.

reference/volumes.html.md.erb
andie787: OK, I'll leave Access a Volume with some links and remove the Mount a Volume section.

PR #719, opened 2023-05-03
troubleshooting - add ways to connect to a service internally and ext…

catflydotio:
…ernally


catflydotio:
To do:

* separate into apps to be reachable by Fly Proxy and apps to be reachable on 6PN
* perhaps move to new services doc in https://github.com/superfly/docs/pull/742
* at least link that doc here
* add all docs links

catflydotio:
Moved content to https://github.com/superfly/docs/pull/742/

PR #730, opened 2023-05-08
Working with machines improvements

magdalene:
I made some improvements while working from the working with machines docs:

* Most importantly: don't send fly API token in header over http when using `api.machines.dev` (instead use https)
* Make the Terraform examples 1% easier to copy-and-paste by using `fly_app`'s `name` attribute instead of making the user update the name in multiple places
* Added a line explaining where to get the auth token when setting up env for machines API

### DIFF
--- a/machines/guides-examples/terraform-machines.html.md.erb
+++ b/machines/guides-examples/terraform-machines.html.md.erb
@@ -80,13 +80,13 @@ resource "fly_app" "exampleApp" {
 }
 
 resource "fly_ip" "exampleIp" {
-  app        = "flyiac" #Replace this with your own app name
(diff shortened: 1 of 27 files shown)


Johannes-Werbrouck (approved):
Hooray for more convenience!

fideloper:
Nice, thanks! 

🤔 One idea to simplify this a bit might be to make the hostname variable include `http://` and `https://` (versus a new variable).

magdalene:
> Nice, thanks!
> 
> 🤔 One idea to simplify this a bit might be to make the hostname variable include `http://` and `https://` (versus a new variable).

Oh yeah, that would be way better. Thanks for the suggestion!

PR #733, opened 2023-05-09
Volume reference re-org

andie787:
An attempt to organize and augment the information about volumes by adding details, using headings, and summarizing and adding to the considerations. Will need a technical review. 

- Moved content out from the Create a volume section into the overview and into separate sections about regions and high availability.
- Added bolded "summary" statements to the bullets in the Considerations section to make them easier to scan visually.
- Created parallel considerations to explain the one-to-one mapping between volumes, regions, apps, and Machines.
- Etc.

### DIFF
--- a/apps/volume-storage.html.md.erb
+++ b/apps/volume-storage.html.md.erb
@@ -10,15 +10,15 @@ order: 30
 
 Volumes are local persistent storage for [Fly Machines](/docs/machines/). Volumes allow an app to save its state&mdash;to preserve configuration, session or user data&mdash;and then be restarted with that information in place. You can access and write to a volume on a Machine just like a regular directory.
 
-**The first rule of Fly Volumes is: Always run at least two volumes per application.** The way to do this is to run at least two Machines per app. Also note that volumes don't sync up by themselves. Your app needs to take care of that. Refer to [Fly Volumes](/docs/reference/volumes/) details about how volumes are implemented.
+**The first rule of Fly Volumes is: Always run at least two volumes per application.** The way to do this is to run at least two Machines per app. Also note that volumes don't sync up by themselves; your app needs to take care of that. Refer to [Fly Volumes](/docs/reference/volumes/) for details about how volumes work.
 
 ## Launch a new app with a Fly Volume
 
-Follow these steps to create a new app, add the volume configuration to `fly.toml`, and then deploy the app. 
+Launch a new app with one Machine and an attached volume, and then clone the Machine to scale out.
 
 ### Launch the app, but don't deploy it yet
 
-The app has to exist for the volume to be created. The volume has to exist before you can mount it to a Machine. The first deployment of your app will create a Machine. So the shortest path to a Machine with a mounted volume begins by launching the app, and saying `N` to "deploy now?":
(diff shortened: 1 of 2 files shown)


reference/volumes.html.md.erb
```diff
@@ -7,46 +7,51 @@ nav: firecracker
 
 <%= partial "/docs/partials/v2_transition_banner" %>
 
-Apps can store ephemeral data on the root file systems of their Machines, but a Machine's file system is rebuilt from scratch each time the app is deployed or the Machine is restarted. Volumes are local persistent storage for [Fly Machines](/docs/machines/). Volumes allow an app to save its state—to preserve configuration, session or user data—and then be restarted with that information in place.
+Apps can store only ephemeral data on the root file systems of their Machines, because a Machine's file system is rebuilt from scratch each time you deploy your app, or the Machine is restarted. Volumes are local persistent storage for [Fly Machines](/docs/machines/). You can use volumes to store your database files or backups, to save your app's state, configuration, and session or user data, or for any information that needs to persist after deploy or restart.
```
catflydotio, line 10: Configuration data, session data and user data are the app's state in the original text. Just had to get that out of my system! 

```suggestion
Apps can store only ephemeral data on the root file systems of their Machines, because a Machine's file system is rebuilt from scratch each time you deploy your app, or the Machine is restarted. Volumes are local persistent storage for [Fly Machines](/docs/machines/). You can use volumes for database storage, for your app's configuration, session or user data, or for any other information that needs to persist after deploy or restart.
```

We don't have a great way to use volumes for backups, do we? You can't write to one without attaching it to a VM, and you can only attach it to one VM at a time.

catflydotio (approved):

apps/volume-storage.html.md.erb
```diff
@@ -10,15 +10,15 @@ order: 30
...
 
 ## Launch a new app with a Fly Volume
 
-Follow these steps to create a new app, add the volume configuration to `fly.toml`, and then deploy the app. 
+Launch a new app with one Machine and an attached volume, and then clone the Machine to scale out.
 
 ### Launch the app, but don't deploy it yet
 
-The app has to exist for the volume to be created. The volume has to exist before you can mount it to a Machine. The first deployment of your app will create a Machine. So the shortest path to a Machine with a mounted volume begins by launching the app, and saying `N` to "deploy now?":
+The app has to exist for the volume to be created. The volume has to exist before you can mount it to a Machine. The shortest path to a Machine with a mounted volume begins by launching the app, and saying `N` to "deploy now?":
```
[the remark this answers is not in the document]
andie787: @catflydotio - Oh interesting! Is that what we'd recommend as a workflow though? Leaving the command line with that question open and then going back and saying `Y`. You'd be opening another terminal window to open the file?

reference/volumes.html.md.erb
andie787: @catflydotio you're right about state. I guess I think of configuration as separate from state, but used together? But I can put it back to the way it was. :)  What do you suggest?

catflydotio: I started bikeshedding on this, but whatever you want. 😆 Volumes are the persistent storage; put data on them that should persist

apps/volume-storage.html.md.erb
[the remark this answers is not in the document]
andie787: Going to leave it as-is for now as three separate steps.

PR #741, opened 2023-05-14
Example: Ping an App via an internal address

nbw:
a step-by-step example using fly internal addresses.

screencapture-localhost-4567-docs-reference-private-networking-2023-05-17-18_17_57

### DIFF
--- a/reference/private-networking.html.md
+++ b/reference/private-networking.html.md
@@ -100,7 +100,7 @@ Examples of retrieving this information are in the [fly-examples/privatenet](htt
 
 ### Example: Ping an app via an `.internal` address
 
-A quick to tutorial to show private networks in action. If you've arrived on private networking, we'll assume you've at least installed the CLI and are at least a little familiar with deploying apps on Fly.
+The following example shows private networks in action. You'll need to install [flyctl](/docs/hands-on/install-flyctl/) and be at least a little familiar with deploying apps on Fly.io.
 
 For this example we'll use the [hello-fly](https://github.com/fly-apps/hello-fly) repository.
 
--- a/reference/private-networking.html.md
+++ b/reference/private-networking.html.md
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
 
 Examples of retrieving this information are in the [fly-examples/privatenet](https://github.com/fly-apps/privatenet) repository.
 
+
+### Example: Ping an app via an `.internal` addresss
+
+A quick to tutorial to show private networks in action. If you've arrived on private networking, we'll assume you've at least installed the CLI and are at least a little familiar with deploying apps on Fly.
+
+For this example we'll use the [hello-fly](https://github.com/fly-apps/hello-fly) repository.
+
+```cmd
+# clone the repository
+git clone https://github.com/fly-apps/hello-fly
+
+cd hello-fly
+
+# create an app on fly
+fly launch
+
+# output omitted...
+
+? Would you like to deploy now? (y/N)
+N
+
+# deploying now is fine, but we'll want to modify
+# the fly.toml and redeploy again
+```
+
+#### Expose an `internal_port` in the fly.toml
+
+In order for internal requests to reach your app, an `internal_port` will need to be exposed in the app's fly.toml that was generated during launch. Here's what the _fly.toml_ might look like exposing port 8080:
+
+```toml
+app = "my-app-name"
+primary_region = "nrt"
+kill_signal = "SIGINT"
+kill_timeout = "5s"
+
+[[services]]
+  protocol = "tcp"
+  internal_port = 8080 # <-- this is the important line
+  processes = ["app"]
+
+# remaining contents omitted
+```
+
+#### Install dig and/or curl in your Dockerfile
+
+To test the internal network we need tool for that. By default dig and curl won't be installed on a machine. Just add `dnsutils` (for dig) and `curl` to your Dockerfile so you can access them later. Here's the line to add to your the Dockerfile:
+
+```docker
+RUN apt-get update -y && \
+  apt-get install -y curl dnsutils \
+  && apt-get clean && rm -f /var/lib/apt/lists/*_*
+```
+
+#### Deploy the app
+
+If you haven't already, run `fly deploy` and make sure the app is running. If you're using the [hello-fly](https://github.com/fly-apps/hello-fly) example then `service.port`'s were set in your fly.toml. You can  visit https://[[ your app name]].fly.dev and see a "Welcome to Fly" message.
+
+#### Ping an internal address
+
+Now it's time for the cool stuff.
+
+1. SSH into the app `fly ssh console -a my-app-name`
+2. Curl the app through the internal network:
+
+```cmd
+curl my-app-name.internal:8080
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Fly.io</title>
+    <link rel='stylesheet' href='/stylesheets/style.css' />
+  </head>
+  <body>
+    <h1>Fly.io (nrt)</h1>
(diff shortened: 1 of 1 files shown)


reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+  <body>
+    <h1>Fly.io (nrt)</h1>
+    <p>Welcome to Fly.io (nrt)</p>
+  </body>
+</html>
+
+# ↑↑↑ the same output as `curl https://my-app-name.fly.dev`
+```
+
+Nice! We've hit our app via the internal network.
```
andie787, line 181:
```suggestion
Nice! You've hit your app via the internal network.
```

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+
```
andie787, line 182:
```suggestion
To see a list of the apps on your private network:
```
suggest moving this line into text above the command

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+```cmd
+# to see a list of apps on your private network:
+dig -t txt _apps.internal +short
+```
+
+#### Tips
+
+- If your app is deployed to multiple regions, you can also hit a different instance of the app based on region using curl [[region]].[[your app name]].internal:8080
+- In the example we pinged the same app we deployed, but try hitting another app in your network from within ssh console
+- Heads up: prefixing an internal network url request with http:// is fine, but obviously there aren't ssl certs issues for an private internal network so https:// won't work
```
andie787, line 192:
```suggestion
- Heads up: prefixing an internal network url request with `http://` is fine, but since ssl certs aren't issued for private internal networks, `https://` won't work
```

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+# output omitted...
+
+? Would you like to deploy now? (y/N)
+N
+
+# deploying now is fine, but we'll want to modify
+# the fly.toml and redeploy again
+```
+
+#### Expose an `internal_port` in the fly.toml
```
andie787, line 125: When I run `fly launch` on `hello-fly` my `fly.toml` looks like this:
```
app = "purple-sound-569"
primary_region = "yyz"

[http_service]
  internal_port = 8080
  force_https = true
  auto_stop_machines = true
  auto_start_machines = true
```
This step should be clear on what needs to be changed in the file, or the example should look like their file?

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+
+```docker
+RUN apt-get update -y && \
+  apt-get install -y curl dnsutils \
+  && apt-get clean && rm -f /var/lib/apt/lists/*_*
+```
+
+#### Deploy the app
+
+If you haven't already, run `fly deploy` and make sure the app is running. If you're using the [hello-fly](https://github.com/fly-apps/hello-fly) example then `service.port`'s were set in your fly.toml. You can  visit https://[[ your app name]].fly.dev and see a "Welcome to Fly" message.
```
andie787, line 155:
```suggestion
If you haven't already, run `fly deploy` and make sure the app is running. If you're using the [hello-fly](https://github.com/fly-apps/hello-fly) example, then `service.port`'s were set in your fly.toml. You can visit https://[[ your app name]].fly.dev and see a "Welcome to Fly.io" message.
```

andie787, line 155: `services-ports` are not set in the default `fly.toml` for apps v2 because it just uses `[http_service]`

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+    <link rel='stylesheet' href='/stylesheets/style.css' />
+  </head>
+  <body>
+    <h1>Fly.io (nrt)</h1>
+    <p>Welcome to Fly.io (nrt)</p>
+  </body>
+</html>
+
+# ↑↑↑ the same output as `curl https://my-app-name.fly.dev`
+```
```
andie787, line 179: split this into command and output:

\```
curl my-app-name.internal:8080
\```

\```output
put output here
\```

andie787 (commented):
This is a great addition to the docs!
Left a few suggestions on formatting, some nitpicks, and a question. :)

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
 
 Examples of retrieving this information are in the [fly-examples/privatenet](https://github.com/fly-apps/privatenet) repository.
 
+
+### Example: Ping an app via an `.internal` address
```
catflydotio, line 101: "Reach" an app? Since you use curl, not ping

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+? Would you like to deploy now? (y/N)
+N
+
+# deploying now is fine, but we'll want to modify
+# the fly.toml and redeploy again
+```
+
+#### Expose an `internal_port` in the fly.toml
+
+In order for internal requests to reach your app, an `internal_port` will need to be exposed in the app's fly.toml that was generated during launch. Here's what the _fly.toml_ might look like exposing port 8080:
```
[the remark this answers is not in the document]
catflydotio: You don't need a service defined at all for internal networking to work. It's only necessary when you're using Fly Proxy for routing, which is the _other_ (i.e. not-on-the-6PN-network) way to reach a Fly.io VM.

reference/private-networking.html.md
```diff
@@ -97,6 +97,101 @@ Finally, You can discover all the apps in the organization by requesting the TXT
...
+Nice! We've hit our app via the internal network.
+
+```cmd
+# to see a list of apps on your private network:
+dig -t txt _apps.internal +short
+```
+
+#### Tips
+
+- If your app is deployed to multiple regions, you can also hit a different instance of the app based on region using curl [[region]].[[your app name]].internal:8080
```
catflydotio, line 190: We use this convention: `<region>.<app-name>.internal:8080`

catflydotio:
Thanks for writing up your exploration! Can you boil it down to the most private-networky information you learned? We probably don't need to start with installing an example app in our private networking docs, but it can be handy to see what happens when you curl an internal address. Can you demystify it a bit? What happens if there's more than one VM it could choose from?

nbw:
@catflydotio I would say the two things that got me lost were:
- adding an `internal_port` in the fly.toml is required  
  - which is why to me seeing the `fly launch` flow is helpful
- installing `curl` and `dig` in the docker file so that it shows up on machines
  - it wasn't obvious that I would add a line to my docker file and it would show up in the machine I `fly ssh console`d into. 

I really wanted a tutorial that doesn't assume too much, but also proves that it works by showing the output as if you're following a developer rather than just saying "you should know linux commands and should see some output". Instead of creating a new app, I could use an existing one I have by making the same modifications (assuming it's a web app with an endpoint that can be hit), but I didn't want to do that or assume that the user is willing to either.

I could see removing this section and/or condensing it (assuming they know what commands to execute instead). the big thing is that they need to have a fly.toml file which is generated during launch:
Screenshot 2023-05-16 at 7 28 58

reference/private-networking.html.md
nbw: i must've been on an old CLI  version, so i've updated my tutorial.

PR #742, opened 2023-05-16
A doc about connecting to services using Fly Proxy and 6PN

catflydotio:
Preview: https://justchecking.fly.dev/docs/getting-started/app-services/

To do:
* proof-read 
* find a better title
* check note numbers XD
* check links
(don't forget about round-robin dns)


getting-started/app-services.html.md
```diff
@@ -0,0 +1,154 @@
...
+#### Private (Flycast) apps
+
+To have Fly Proxy load-balance among VMs on a non-public app, [allocate a _private_ (Flycast) IPv6 address](/docs/reference/private-networking/#assigning-a-flycast-address) using `fly ips allocate-v6 --private`, and **remove any public IPs from the app** using `fly ips release`. A Flycast IP can only be reached from within the private network it's allocated on, which must belong to the same Fly Organization as the Fly App it points to. 
+
+<section class="warning icon">
+If your configuration includes any services for Fly Proxy to route to, and the app has a public IP, that service is exposed to the whole internet.</section>
+
+### A note on Fly App IPs and Fly-Replay routing
+
+App-wide IP addresses (public and private) tell Fly Proxy which app to deliver a request to. If a service _only_ needs to be reachable by Fly Proxy in its handling of the [Fly-Replay](/docs/reference/dynamic-request-routing/) response header, its app does not need an IP; app information is incorporated directly in the header.
```
andie787, line 63:
```suggestion
App-wide IP addresses (public and private) tell Fly Proxy which app to deliver a request to. If a service _only_ needs to be reachable by Fly Proxy in its handling of the [Fly-Replay](/docs/reference/dynamic-request-routing/) response header, then its app does not need an IP; app information is incorporated directly in the header.
```

getting-started/app-services.html.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Private services on 6PN (direct WireGuard connections)
+
+If you don't need load balancing or other Fly Proxy features, you (or one of your apps) can connect directly on a VM's 6PN address: its address on its IPv6 private WireGuard network.
```
andie787, line 67:
```suggestion
If you don't need load balancing or other Fly Proxy features, then you (or one of your apps) can connect directly on a VM's 6PN address, which is the VM's address on its IPv6 private WireGuard network.
```
not sure about this one, the last but could also be in parentheses

getting-started/app-services.html.md
```diff
@@ -0,0 +1,154 @@
...
+
+That's it, really. There's no need to define any service in `fly.toml`, and there's no need to provision any IPs for the app.
+
+If your service works publicly but can't be reached directly over WireGuard, the problem may be that it's not listening on IPv6.
+
+### Fly.io internal DNS and 6PN addresses 
+
+Fly.io runs a specialized internal DNS service to find 6PN addresses for VMs. Queries on the various `.internal` names return either AAAA records for 6PN addresses, or a TXT record with information about your app, your network, or your Machines. 
+
+Internal DNS gives us some useful presets for targeting requests. For instance, a DNS query on the address `<region>.<appname>.internal` yields the 6PN addresses of the VMs of app `<appname>` in region `<region>`. Addressing a request to `<region>.<appname>.internal` gets it delivered to one of those VMs.
```
andie787, line 83:
```suggestion
Internal DNS can give you some useful presets for targeting requests. For instance, a DNS query on the address `<region>.<appname>.internal` yields the 6PN addresses of the VMs of app `<appname>` in region `<region>`. Addressing a request to `<region>.<appname>.internal` gets it delivered to one of those VMs.
```

getting-started/app-services.html.md
```diff
@@ -0,0 +1,154 @@
...
+
+Our [private networking docs](https://fly.io/docs/reference/private-networking/#fly-internal-addresses) list the `.internal` names and what kind of information a DNS query returns for each.
+
+## A note on IPv4 and IPv6 wildcards
+
+Services to be routed to using Fly Proxy need to bind to IPv4 addresses, and services to be reached over 6PN need to bind to IPv6, as described above. However, you may find that your service works even if you haven't used the addresses we specified. IPv4-mapped IPv6 addresses, if enabled on the VM, can allow IPv4 (and thus Fly Proxy) connections to a service bound on `[::]`. Further, the wildcard syntax `[::]` or `0.0.0.0` sometimes covers both IPv4 and IPv6, depending on the language, library, or application.
```
andie787, line 89:
```suggestion
Services to be routed to using Fly Proxy need to bind to IPv4 addresses, and services to be reached over 6PN need to bind to IPv6, as described above. However, you may find that your service works even if you haven't used the addresses we specified. IPv4-mapped IPv6 addresses, if enabled on the VM, can allow IPv4 (and thus Fly Proxy) connections to a service bound on `[::]`. Further, the wildcard syntax, `[::]` or `0.0.0.0`, sometimes covers both IPv4 and IPv6, depending on the language, library, or application. 
```

andie787 (commented):
Nice! Left some nitpicky suggestion for ya.
Assuming you are using "VM" vs "Machine" here on purpose.

getting-started/app-services.html.md
catflydotio: > Nice! Left some nitpicky suggestion for ya. Assuming you are using "VM" vs "Machine" here on purpose.

Yes, for now.

getting-started/app-services.html.md
catflydotio: Disagree. 😄 The presets _do_ exist; we _can_ use them.

getting-started/app-services.html.md
```diff
@@ -0,0 +1,154 @@
...
+title: Connect to an App Service
+layout: docs
+sitemap: false
+nav: firecracker
+---
+
+There are two basic ways to talk to a process running in your Fly.io VM: 
+
+1. Via Fly Proxy, the Fly.io component that handles load balancing&mdash;this is what you'll need for any public web service
+2. Over the WireGuard [IPv6 private network ("6PN")](/docs/reference/private-networking/) that the app belongs to&mdash;this can be useful for, e.g., providing private supporting services to your Fly Apps
```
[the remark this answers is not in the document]
catflydotio: I wouldn't like to limit people's imaginations. It's an example. I think e.g. is the clearest, most concise way to say this.

getting-started/app-services.html.md
catflydotio: I will resist adding "then" everywhere I use "if" (I had to hold myself back from removing one of yours yesterday 😆)

getting-started/app-services.html.md
catflydotio: I added parentheses.

PR #744, opened 2023-05-16
removed `--force-[nomad/machines]` and `--[nomad/machines]` flags

alichay:
docs changes for https://github.com/superfly/flyctl/pull/2287


catflydotio:
This is great while our auto docs generation is awol! ✅

PR #752, opened 2023-05-19
Suggest pwsh instead of powershell

haileys:
`powershell.exe` is fixed at PowerShell 5.1 for backwards compatibility reasons, newer versions use `pwsh.exe`.

Older versions have [issues](https://github.com/superfly/flyctl/pull/2307), we should suggest users to use a newer PowerShell instead.

### DIFF
--- a/hands-on/install-flyctl.html.md.erb
+++ b/hands-on/install-flyctl.html.md.erb
@@ -58,7 +58,7 @@ curl -L https://fly.io/install.sh | sh
 Run the Powershell install script:
 
 ```cmd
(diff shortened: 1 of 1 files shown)


haileys:
@catflydotio Good catch. I've added an extra piece saying that the `powershell` command can be used instead if `pwsh` is not found, but also recommending that the user install the latest version of PowerShell

haileys:
I noticed that we write `Powershell` in our docs, though it is officially stylised as `PowerShell`. I've opted to stay consistent with the rest of our docs here. Maybe we want to update the rest of them, I'm not too fussed either way.

PR #767, opened 2023-05-25
Remove LiteFS banner; replace with project status disclaimer

benbjohnson:
Previously, there was a banner on every page noting that LiteFS was beta and not ready for production. That is probably overly cautious advice at this point. LiteFS is stable and is in production deployments.

Instead, a "project status" section has been added to the home page of the LiteFS docs to note its pre-1.0 status and also recommend that people use a cluster backup strategy. That's good advice for any database, honestly.

/cc @kentcdodds

### DIFF
--- a/litefs/backup.html.md.erb
+++ b/litefs/backup.html.md.erb
@@ -1,0 +1,51 @@
+---
+title: Backing up your LiteFS cluster
+layout: docs
+sitemap: false
+nav: litefs
+toc: true
+---
+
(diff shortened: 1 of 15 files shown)


kentcdodds:
*THANK YOU* 😆

benbjohnson:
@kentcdodds It takes a couple minutes for the docs to go through CI and deploy live but you should see it soon.

PR #768, opened 2023-05-26
Remove redundant wording

acoffman:
The "M" in RAM already stands for memory. 

Love Fly, and love your extensive docs. Thanks for all the work you put in on them!


benwaffle:
Thanks!

PR #770, opened 2023-05-26
Update top level checks docs for clarity and include the missing headers field

jsierles:


andie787:
Perfect! Thanks for catching the `processes` being in the wrong place too :)

PR #783, opened 2023-06-01
Fix services.concurrency configuration example

szepeviktor:
services.concurrency seems to have one pair of brackets.


wjordan:
Although `[[services]]` is an [array of tables](https://toml.io/en/v1.0.0#array-of-tables), `[services.concurrency]` itself is a sub-[table](https://toml.io/en/v1.0.0#table) (not a sub-array of tables), so it uses single brackets.

szepeviktor:
Thank you!!

PR #794, opened 2023-06-02
Restored the alphabetical order of the commands in the nav on https://fly.io/docs/flyctl/

Johannes-Werbrouck:


Johannes-Werbrouck:
The commands in the nav on the [flyctl reference](/fly.io/docs/flyctl/) were out of order: 
Scherm­afbeelding 2023-06-02 om 11 24 15

catflydotio:
We put `launch`, `apps` and `machines` at the top because we wanted to highlight them. Knowing that, is there still a change you would want to make?

Johannes-Werbrouck:
> We put `launch`, `apps` and `machines` at the top because we wanted to highlight them. Knowing that, is there still a change you would want to make?

I had no clue these were highlighted, it just seemed out of order. I'd re-order them like this then: apps - launch - machines - auth - certs - rest of commands. This way, there's a clear alphabetical order for the important and other commands. 

I'd also make it extra clear these commands are highlighted, I don't know how feasible it is to split up the list of commands into 'important' and 'other' ?

PR #796, opened 2023-06-02
Resilient apps doc

andie787:
A doc to bring together and explain some of the features we have to make apps more resilient. 

Preview: https://aa-docs.fly.dev/docs/reference/app-availability/


reference/high-availability-apps.html.md.erb
```diff
@@ -0,0 +1,121 @@
...
+
+* creates and starts [two Machines](#two-machines-for-process-groups-with-services) for process groups with services, and sets automatic start and stop to true and minimum machines running to zero
+* creates and starts one Machine and creates one stopped [standby Machine](#standby-machines-for-process-groups-without-services) for process groups without services
+* creates and starts only one Machine if the process group has volumes mounted
+
+### Two Machines for process groups with services
+
+The most basic way to ensure high availability is to create more than one Machine per process group.
+
+If a process group has services defined, two Machines are automatically created and started when:
```
fideloper, line 59: It might be worth being explicit that this means people's apps will be load balanced by default, since that comes with a slew of concerns like "session storage" (to ensure people stay logged in when they're bounced between different servers), and "where do my files go?" - which is covered under the Volumes section, but might not be obvious that by default you need to care about that now.

This grouchy comment is brought to you by needing to adjust `fly launch` for Laravel apps because of session storage in load balanced environments 😂

fideloper (commented):
So much good info is crammed in here, nice job!

I left some feedback, feel free to ignore literally any/all of it as you see fit.

reference/high-availability-apps.html.md.erb
```diff
@@ -0,0 +1,121 @@
...
+## Fly.io high availability features
+
+The Fly Apps V2 platform has features to help make your app more resilient in case of outages or hardware failures.
+
+<div class="callout">
+<b>A note about Fly Proxy</b>: 
+Fly Proxy does a lot of work in the background, applying handlers, managing traffic and load balancing, and managing connections to services. Fly Proxy also automatically starts and stops Machines based on traffic to your app. The proxy works for [apps or processes with services defined](/docs/getting-started/app-services/#services-routed-with-fly-proxy).
+</div>
+
+The features for redundancy and availability are:
```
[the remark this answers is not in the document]
andie787: Agreed! Adding anchor links where I can

reference/high-availability-apps.html.md.erb
```diff
@@ -0,0 +1,121 @@
...
+
+These features require or depend on your app configuration in the fly.toml file. Learn more about [app configuration](/docs/reference/configuration/).
+
+You might also be interested in how we do [load balancing](/docs/reference/load-balancing/).
+
+## High availability by default on first deploy
+
+The first time you deploy your app to the Fly.io platform, you get some default high availability configurations. These settings help you to deploy a recommended setup without having to think about it. On the other hand, if the defaults aren't what you're looking for, then you can still configure your app and processes the way you want. Learn more about [scaling Machine CPU and RAM](/docs/apps/scale-machine/) and [scaling the number of Machines](/docs/apps/scale-count/).
+
+A Fly App with no extra process groups defined still uses the `app` process group by default. So there's no such thing as an app with no process groups.
```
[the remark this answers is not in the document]
andie787: Was also thinking this. I will see how it feels there, or somewhere else, or as a note with more info in the "worst case" :)

reference/high-availability-apps.html.md.erb
andie787: @fideloper Maybe this info belongs in "What your app needs to do"? Then I'd probably need to move the note about Fly Proxy up into that section too... 🤔

reference/app-availability.html.md.erb
```diff
@@ -0,0 +1,124 @@
...
+<p><b>A note about Fly Proxy</b>: 
+Fly Proxy does a lot of work in the background, applying handlers, managing traffic and load balancing, and managing connections to services. Fly Proxy also automatically starts and stops Machines based on traffic to your app. The proxy works for [process groups with services defined](/docs/getting-started/app-services/#services-routed-with-fly-proxy).
+</div>
+
+Since Fly Proxy does load balancing by default, you need to consider how your app handles:
+
+* session storage, so that users stay logged in when their session moves to a different server
+* persistent file storage and replication, as discussed in the following paragraphs
+
+For databases, you'll need to follow the recommendations specific to your deployment, including how to set up clusters for replication and failover.
```
catflydotio, line 24: I don't know if it's just me, but my brain is sliding over "your deployment" as having indeterminate meaning--reading the end of the sentence mitigates that, but I'm lost for a second. Is it recommendations specific to the database? Specific to the desired cluster config?

reference/app-availability.html.md.erb
```diff
@@ -0,0 +1,124 @@
...
+
+If you're using Fly Postgres, our un-managed database, then you need to [manage backups, recovery, replication, and scaling](/docs/postgres/getting-started/what-you-should-know/) yourself. For information about high availability with Fly Postgres, refer to [High Availability & Global Replication](/docs/postgres/advanced-guides/high-availability-and-global-replication/).
```
catflydotio, line 26:
```suggestion
If you're deploying Fly Postgres, our un-managed database app, then you need to [manage backups, recovery, replication, and scaling](/docs/postgres/getting-started/what-you-should-know/) yourself. For information about high availability with Fly Postgres, refer to [High Availability & Global Replication](/docs/postgres/advanced-guides/high-availability-and-global-replication/).
```

reference/app-availability.html.md.erb
```diff
@@ -0,0 +1,124 @@
...
+
+When you use [Fly volumes](/docs/reference/volumes/) with your app, your app needs to implement replication and backups. Volumes are block devices that live on a single disk array and get mounted directly by your Machine. There's no magic. And because there's no built-in replication between volumes, Fly.io never automatically starts multiple Machines with volumes on the first deploy. If your app provides a clustering mechanism with data replication, like most databases do, we recommend you take advantage of that and use multiple volumes per app. We place your app's volumes in different hardware zones within a region to mitigate hardware failures.
```
catflydotio, line 28: "Fly Volumes"

catflydotio: > And because there's no built-in replication between volumes, Fly.io never automatically starts multiple Machines with volumes on the first deploy.

I'd leave this point until after the feature is introduced! `fly deploy` is the only thing that'll automatically create an extra Machine and it's super confusing outside that context.

catflydotio: > If your app provides a clustering mechanism with data replication, like most databases do, we recommend you take advantage of that and use multiple volumes per app. 

What about "we recommend that you take advantage of that and run multiple instances (each with its own attached volume)"? You can't run a volume without a VM, but people don't always get that off the bat.

catflydotio: > We place your app's volumes in different hardware zones within a region to mitigate hardware failures.

"When possible, we place your app's volumes in different hardware zones within a region to mitigate the impact of potential hardware failures." ?

reference/app-availability.html.md.erb
```diff
@@ -0,0 +1,124 @@
...
+The Fly Apps V2 platform has features to help make your app more resilient in case of outages or hardware failures.
+
+The features for redundancy and availability are:
+
+* [Redundancy by default on first deploy](#redundancy-by-default-on-first-deploy): standby Machines and two Machines per process group
+* [Automatically start and stop Machines](#automatically-start-and-stop-machines) based on traffic
+* [Health check-based routing](#health-check-based-routing)
+
+<div class="callout">
+These features require or depend on your app configuration in the `fly.toml` file, and this includes process groups. Every Fly app has at least one process group. An app with no extra process groups defined in `fly.toml` still uses the `app` process group by default. Learn more about [app configuration](/docs/reference/configuration/).
```
catflydotio, line 43:
```suggestion
These features require or depend on your app configuration in the `fly.toml` file, and this includes process groups. Every Fly App has at least one process group. An app with no extra process groups defined in `fly.toml` still uses the `app` process group by default. Learn more about [app configuration](/docs/reference/configuration/).
```

reference/app-availability.html.md.erb
```diff
@@ -0,0 +1,124 @@
...
+A resilient app is resistant to events like hardware failures or outages. Redundancy, supported by monitoring and scaling, is one of the ways to make your app more resilient, and we provide some built-in features to make this easier.
+
+But first, it's important to understand what your app needs to take care of.
+
+## What your app needs to do
+
+<div class="callout">
+<p><b>A note about Fly Proxy</b>: 
+Fly Proxy does a lot of work in the background, applying handlers, managing traffic and load balancing, and managing connections to services. Fly Proxy also automatically starts and stops Machines based on traffic to your app. The proxy works for [process groups with services defined](/docs/getting-started/app-services/#services-routed-with-fly-proxy).
+</div>
```
Johannes-Werbrouck, line 17: A callout to me feels like a deeper dive, something extra that needs to be explained. That's why I'd suggest moving the Fly Proxy callout to L23. This way, you talk about the Fly Proxy stuff that's directly related to this doc and after that show the callout that explains the Fly Proxy further. This is very much an opinion so feel free to ignore.

andie787: @Johannes-Werbrouck thanks for this. I've moved this bit around a few times! Wasn't entirely happy with where it ended up. :) I'll give your idea a try.

PR #800, opened 2023-06-05
Edits for autostop and apps that exit from within when idle

andie787:
Edits based on some new (to me) info about not using the autostop feature when your app exits on it's own when it's idle for a period. Source: https://community.fly.io/t/automatically-starting-stopping-apps-v2-instances/12342/46


apps/autostart-stop.html.md.erb
```diff
@@ -44,16 +46,16 @@ Default settings in `fly.toml` for V2 apps:
 ```
 
 <div class="callout">
-Existing V2 apps—or any V2 apps that don't have these settings in `fly.toml`—will automatically start but not automatically stop Machines.
+Existing V2 apps—or any V2 apps that don't have these settings in `fly.toml`—have the default values `auto_start_machines = true` and `auto_stop_machines = false`.
```
catflydotio, line 49:
```suggestion
Existing V2 apps&mdash;or any V2 apps that don't have these settings in `fly.toml`&mdash;have the default values `auto_start_machines = true` and `auto_stop_machines = false`.
```

PR #809, opened 2023-06-09
Litefs cloud docs

magdalene:
This is SUPER SUPER WIP. 

I've added some LiteFS cloud docs. Some notes:

* I think we should rethink the getting started page (as we discussed, Ben), but I would leave that for a different PR. The goal here is to integrate LiteFS Cloud info into the existing docs as they are.
* I've written the docs from a web UI first perspective. It might happen that we release without the web UI ready yet, and obviously that would require changes here, but I want to see the docs more how they "should be" first?
* Some of the UI things (both web & flyctl) I've written about don't exist yet, so details may change obviously! 

Here's what kind of feedback is most useful at this stage:

* Feedback about the UX implied/described by the docs. (Should we be / are we building something different than the docs say.)
* Feedback about the high level structure or focus of the docs.
* Feedback about big topics that are missing or wrong.

(Other feedback is also welcome of course, but not the main focus right now!)

Links to preview of docs that are new/changed here:

* [Getting started](https://litefs-cloud-docs.fly.dev/docs/litefs/getting-started/)
* [Using LiteFS Cloud for backups](https://litefs-cloud-docs.fly.dev/docs/litefs/cloud-backups/)
* [Configuring LiteFS Cloud with flyctl](https://litefs-cloud-docs.fly.dev/docs/litefs/cloud-flyctl/)
* [Backing up your cluster](https://litefs-cloud-docs.fly.dev/docs/litefs/backup/) (minor update to link to the doc about backing up with litefs cloud)


litefs/cloud-flyctl.html.md.erb
```diff
@@ -0,0 +1,35 @@
...
+your LiteFS cluster. Your data will continue to be available as long as
+your LiteFS cluster has nodes. We recommend to
+[configure alternative backups][Backing up your LiteFS cluster] before taking
+this action.
+
+[Backing up your LiteFS cluster]: /docs/litefs/backup
+
+```cmd
+fly litefs-cloud delete $(cluster name)
+```
```
benbjohnson, line 35: The deletion feature is non-existent, however, once we do have it we should instruct people to remove the `LITEFS_CLOUD_TOKEN` and restart their LiteFS nodes before deleting. Otherwise they'll get a bunch of errors in their logs saying the cluster doesn't exist.

litefs/cloud-flyctl.html.md.erb
```diff
@@ -0,0 +1,35 @@
...
+
+```cmd
+fly litefs-cloud create
+```
+
+## Get a LiteFS Cloud token
+
+```cmd
+fly tokens create litefs-cloud --cluster $(cluster name)
+```
```
benbjohnson, line 21: I'm not sure how we do user data in other parts of our docs. I wonder if the `$(cluster name)` might look confusing? I'd personally go with just a `my-cluster` name to indicate that the user should replace it with their cluster name.

benbjohnson (commented):
This is all looking great! I added a handful of comments but overall I think you're going in the right direction.

litefs/cloud-flyctl.html.md.erb
magdalene: As far as I've seen, we do both things in the docs (there doesn't appear to be a convention). I can switch to `my-cluster` instead!

magdalene:
I'm going to close this PR in favor of https://github.com/superfly/docs/pull/836 (which has all the changes from here, plus some other improvements).

Issue #804, opened 2023-06-06
Broken link 


arjunan-r:
This link https://fly.io/docs/going-to-production/monitoring/exporting-logs/nats.io 
on this page https://fly.io/docs/going-to-production/monitoring/exporting-logs/#internals
requires to be updated to https://nats.io/)

catflydotio:
Fixed by #813

PR #813, opened 2023-06-10
Incorrect link r|nats.io|https://nats.io

DazWilkin:
Incorrect link for NATS.

Was `nats.io` which results in a relative (i.e. `https://fly.io/docs/going-to-production/monitoring/exporting-logs/nats.io`) reference rather than `https://nats.io`


catflydotio:
Fixes #804.

Thanks @DazWilkin

PR #816, opened 2023-06-11
Include a Laravel-Inertia Server Side Render Support Deploy to Flyio Section at Laravel Docs

KTanAug21:


KTanAug21:
Got a community question on how to deploy a  Inertia-SSR app in Fly.io! This will help future inertia-ssr devs easily deploy their apps to fly.io

laravel/advanced-guides/using-inertia-ssr.html.md
```diff
@@ -0,0 +1,46 @@
...
+[http_service]
+  internal_port = 8080
+  force_https = true
+  auto_stop_machines = true
+  auto_start_machines = true
+  min_machines_running = 0
++  processes = ["app"]
+```
+
+With that, do a quick `fly deploy` and your good to go!
```
Johannes-Werbrouck, line 46: your -> you're

Johannes-Werbrouck (commented):
Great work, lots of people use inertia so I'm guessing this will be a successful post!

laravel/advanced-guides/using-inertia-ssr.html.md
KTanAug21: Thank you @Johannes-Werbrouck ! I'll see if I can create a post/article of this as well

fideloper:
> The Laravel fly-scanner already generates a Dockerfile that makes use of a Node image. So we can simply revise that Dockerfile to copy over the Node image's node binaries, with your app's generated node_modules to the final fly-laravel image:

So I'm not totally sure this is enough, I'd be pretty suspicious that there's like  underlying C libraries missing from the container that could cause random errors for people. But I also don't know for sure, right now it's more of a suspicion.

fideloper:
```toml
[processes]
  app=""
  ssr="php /var/www/html/artisan inertia:start-ssr"
```

Does this definitely work? IIRC the backend (PHP) hits the SSR request when a "direct" HTTP request comes in (usually the first request that comes into the server, before a client is loaded that can do the remaining work via AJAX).

PHP needs to know where to make a request against the SSR binary over the network. I'm not sure how it would just know where it's located when it's on a separate VM.

PR #828, opened 2023-06-15
Add `exec` to Experimental section of App Configuration docs

tedmiston:
Adds the `exec` section to the Experimental section of the Configuration docs.

[The other 2 flags used in the code snippet are documented already, but this one is currently missing.]

https://fly.io/docs/reference/configuration/#the-experimental-section

PR #829, opened 2023-06-15
Add experimental `exec` override to docs

tedmiston:
Adds the `exec` section to the Experimental section of the Configuration docs.

[The other 2 flags used in the code snippet are documented already, but this one is currently missing.]

https://fly.io/docs/reference/configuration/#the-experimental-section

(Previously #828 but accidentally closed it when I deleted my fork.)

### DIFF
--- a/reference/configuration.html.md.erb
+++ b/reference/configuration.html.md.erb
@@ -618,3 +618,7 @@ This overrides the `CMD` set by the Dockerfile. It should be specified as an arr
 ### `entrypoint`
 
 This overrides the `ENTRYPOINT` set by the Dockerfile. It should be specified as an array of strings, as seen in the example above.
+
+### `exec`
+
+This overrides the `EXEC` set by the Dockerfile. It should be specified as an array of strings, as seen in the example above.


andie787:
You are right that `exec` is missing from the docs. But it looks like the description should be:

> This overrides both the CMD and ENTRYPOINT set by the Dockerfile. It should be specified as an array of strings, as seen in the example above.

@dangra Should we add this? I can see it is implemented in V2.

reference/configuration.html.md.erb
```diff
@@ -618,3 +618,7 @@ This overrides the `CMD` set by the Dockerfile. It should be specified as an arr
 ### `entrypoint`
 
 This overrides the `ENTRYPOINT` set by the Dockerfile. It should be specified as an array of strings, as seen in the example above.
+
+### `exec`
+
+This overrides the `EXEC` set by the Dockerfile. It should be specified as an array of strings, as seen in the example above.
```
dangra, line 624: There is no "EXEC" statement in Dockerfile. 

If set, the `exec` fly.toml directive overrides both, the ENTRYPOINT and CMD.

dangra:
@andie787 I think se should, it is part of v2 but the suggested description is wrong.

andie787:
Yep, that's similar to what I suggest in my first comment 😉 

Thanks!

andie787:
Replaced by #953

PR #838, opened 2023-06-22
Updates here and there for some clarification about regions

andie787:
This PR attempts to clarify some text about regions and how to add and scale into them. There are also some minor edits done along the way, like fixing some title case in files I was looking at. :)

### DIFF
--- a/apps/scale-count.html.md.erb
+++ b/apps/scale-count.html.md.erb
@@ -48,7 +48,7 @@ other   3       shared          4       1024 MB ord(2),scl
 task    4       performance     1       2048 MB iad,mia,ord,scl
 ```
 
-This app has 10 Machines total. The `app` process group has two in scl (Santiago) and one in mia (Miami). `other` has two in ord (Chicago) and one in scl. `task` has one Machine in each of iad (Ashburn, Virginia), mia, ord, and scl. You can put your Machines wherever works best for your app!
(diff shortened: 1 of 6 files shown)


apps/scale-count.html.md.erb
```diff
@@ -87,27 +87,29 @@ Now there are four Machines in yyz.
...
 
-The `--region` option takes one or more [region codes](/docs/reference/regions/). `fly scale count` creates and/or destroys Machines to reach the specified target count across the regions you list in this option, and tries to balance its changes across these regions. 
+The `--region` option takes one or more [region codes](/docs/reference/regions/). `fly scale count` creates and/or destroys Machines to reach the specified target count across the regions you list in this option, and tries to balance its changes across these regions.
 
-If `--region` is not used, the target count is distributed across all regions in which the app already has platform Machines belonging to any process group.
-
-You can set the scale explicitly per region by specifying a single region per `fly scale` command.
-
-An example: to end up with a total of three Machines between the yyz and ewr (Secaucus, NJ) regions on an app:
+For example, to end up with a total of three Machines between the yyz and ewr regions on an app:
```
catflydotio, line 92: Generally I include the city name when we first mention a region in a doc. Is this an indication you'd like us not to do that?

andie787: Oh I'm fine with it if that's what you do. I was going for consistency within the sentence I guess. I'm not sure if people read the whole doc in order. Is it a convention we want everywhere in docs? @catflydotio

catflydotio: We could go with once per section to be more consistent. It feels like a lot, but I'm inclined to think that on a scale from zero times in the page to every time a region's mentioned, zero times is the thing that's most likely to cause confusion or frustration. If you have an opinion, let's go with your opinion.

I see that Toronto is already written out more than once in the page.

PR #840, opened 2023-06-22
Update docs for bluegreen deploy strategy in V2 apps

andie787:
This PR updates the docs to remove the limitation of no bluegreen deploy strategy for V2 apps.

Source: https://github.com/superfly/flyctl/pull/2458

### DIFF
--- a/apps/deploy.html.md.erb
+++ b/apps/deploy.html.md.erb
@@ -69,7 +69,7 @@ You can specify one of the following deployment strategies:
 * `rolling` (default): wait for each Machine to be successfully deployed before starting the update of the next one
 * `immediate`: bring all Machines down for update at once
 * `canary`: boot a single new Machine, verify its health, and then proceed with a rolling restart strategy
-* `bluegreen`: boot a new Machine alongside each running Machine in the same region, and migrate traffic to the new Machines only once they pass health checks
+* `bluegreen`: boot a new Machine alongside each running Machine in the same region, and migrate traffic to the new Machines only once all the new Machines pass health checks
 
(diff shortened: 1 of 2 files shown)


apps/deploy.html.md.erb
```diff
@@ -69,6 +69,10 @@ You can specify one of the following deployment strategies:
 * `rolling` (default): wait for each Machine to be successfully deployed before starting the update of the next one
 * `immediate`: bring all Machines down for update at once
 * `canary`: boot a single new Machine, verify its health, and then proceed with a rolling restart strategy
+* `bluegreen`: boot a new Machine alongside each running Machine in the same region, and migrate traffic to the new Machines only once they pass health checks
```
catflydotio, line 72:
```suggestion
* `bluegreen`: boot a new Machine alongside each running Machine in the same region, and migrate traffic to the new Machines only once all new Machines pass health checks
```

I wonder what happens if they don't?

catflydotio: For all of these, they only have health checks if the user defines some, right? I wonder what happens if you don't set checks and the app crashes on the new Machines.

andie787: maybe @gwuah can help answer these questons! 🙂

andie787: @gwuah which health checks does it use? Is it only the top level ones?

andie787: I'm going to merge this rather than hold it up for new info about health checks that we can add later. I think the health checks failing, as well as which health checks are used applies to all the strategies with the exception of `immediate`.

andie787: No worries! It should be another PR anyway. :) TY!

PR #842, opened 2023-06-23
Update index.html.md.erb with flyctl authentication step

awebartisan:
Although this step is done automatically the first time we run the `fly` command, I think it should be documented here as part of the process.


andie787:
Thanks for the suggestion @awebartisan. The authentication step is covered in this doc by the content of [this partial](https://github.com/superfly/docs/blob/main/languages-and-frameworks/partials/_flyctl.html.erb) which links to this page: https://fly.io/docs/getting-started/log-in-to-fly/#first-time-or-no-fly-account-sign-up-for-fly.

PR #844, opened 2023-06-23
Document flyctl's secrets heuristics

billyb2:


andie787 (commented):
I hate to discourage documentation, but I'm wondering if this needs to be in docs, since we don't guarantee that we'll catch all potential secrets in env variables. And your warning message is nice and clear if someone does get it.

billyb2:
No that's totally understandable. I'll just keep it in flyctl

andie787:
https://github.com/superfly/flyctl/pull/2470

PR #846, opened 2023-06-26
Update continuous-deployment-with-github-actions.html.md

peteretelej:
minor edit, go example app no longer uses a Dockerfile. this updates outdated doc that says it does.


andie787:
Nice catch! Thanks

PR #851, opened 2023-06-28
Mention platform region command when creating a volume

anthonypz:
When creating a volume, the docs mention the region option (e.g. `fly volumes create myapp_data -r lhr`) multiple times. Although the region option can be found on the [volumes-create page](https://fly.io/docs/flyctl/volumes-create/), this change makes it easier for users to quickly look up their nearest region and associated region code name by mentioning the `fly platform regions` command up front.

### DIFF
--- a/apps/volume-storage.html.md.erb
+++ b/apps/volume-storage.html.md.erb
@@ -137,7 +137,7 @@ destination="/data"
 
 ### Create the volume
 
-Create the volume (or volumes) in the same region as your app.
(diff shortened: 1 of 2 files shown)


andie787:
thanks for this suggestion @anthonypz 
Since we use the region flag for many commands throughout the docs, it might not be practical to add this info everywhere. But I can see the need, so we'll look at adding the `fly platform regions` command in some places and/or linking to the [regions](https://fly.io/docs/reference/regions/) reference page.

PR #856, opened 2023-06-28
Update troubleshooting.html.md: fly vm replaced by fly machine

oatnog:
Docs refer to legacy `fly vm` command. As flyctl cli suggests, change this to `fly machine`


andie787:
Thank you! You're right. This will be fixed by a bunch of updates to the doc in this PR: #891

PR #860, opened 2023-06-30
Fixed link to Litefs Cloud in Cloud Backup section

gautema:
The link in the docs failed. I updated with new url


magdalene (approved):
Thanks so much for catching this, and fixing it!

PR #878, opened 2023-07-06
Guide: Run a SvelteKit App

MarcoCestari:
Observing the community forum, it appears that there is a demand for a SvelteKit Guide. Here is a straightforward guide to meet this demand.


languages-and-frameworks/svelte.html.md
```diff
@@ -0,0 +1,223 @@
...
+		adapter: adapter()
+	}
+};
+
+export default config;
+```
+
+## _Configuring the `start` command_
+
+The `start` command is used to start your application, and you will need it before you deploy. Add the following line to your `package.json` file:
```
rubys: This says "Add the following line", then proceeds to show three lines.  What would be better is to say something along the lines of "find the line containing `"scripts": {` and add the following line after it.  And end that line with a comma.

rubys: I have a suggestion, but I don't want to hold up the merge of this guide before that suggestion is explored and resolved.

My ideal is that `fly launch` will make suggestions and even apply the changes.  Currently the node framework scanner requires a start script - that could be changed to either a start script OR svelte is found in package.json.  This would allow progress to proceed to the @flydotio/dockerfile-node package.  That package could detect svelte, install the node adapter, modify `svelte.config.js`, and add the script to `package.json`.

languages-and-frameworks/svelte.html.md
```diff
@@ -0,0 +1,223 @@
...
+...
+```
+
+Now, let's run `npm install` to install the necessary dependencies and proceed with configuring the adapter.
+
+## _Configuring the Adapter_
+
+Before you can deploy your SvelteKit app, you need to _adapt_ it for your deployment target. Adapters are small plugins that take the built app as input and generate output for deployment.
+
+In this example, to deploy your app on Fly.io, we'll install the Node Adapter. You can install it by executing the command `npm i -D @sveltejs/adapter-node` and then adding it to our `svelte.config.js` file. Here's all the code:
```
rubys: When I ran this, I found that there was already a `svelte.config.js` file present.  I then mentally diffed what I saw there with what you provided below and came to the conclusion that all I needed to modify was the import line.  Let's not make people repeat that process, and instead tell them the one line they need to modify is the import statement.  Since the file is small, it is fine for the guide to show the resulting file.

languages-and-frameworks/svelte.html.md
MarcoCestari: Fixed, please let me know if you have any additional suggestions on how we can further improve it.

languages-and-frameworks/svelte.html.md
MarcoCestari: Fixed

languages-and-frameworks/svelte.html.md
MarcoCestari: Certainly, it would be better to eliminate this task for the user. Another option is to create a specific adapter for Fly.io, which would be automatically recognized by [`adapter-auto`](https://kit.svelte.dev/docs/adapter-auto). This adapter would use `adapter-node` and add the necessary `start` command to `package.json`.

I can further explore the implementation of this at a later time, once we have identified the best solution.

rubys:
Thanks!

PR #882, opened 2023-07-10
Add disaster recovery documentation for litefs

magdalene:
I've added a guide to describe how to recover data from LiteFS Cloud in case you lose all the nodes in your cluster.

### DIFF
--- a/litefs/disaster-recovery.html.md.erb
+++ b/litefs/disaster-recovery.html.md.erb
@@ -82,7 +82,7 @@ Then update the `clusterid` file:
 echo LFSC25AD000F32ED02A9 > /var/lib/litefs/clusterid
 ```
 
-After this, you should restart litefs. If you're using the Fly Platform, close
+After this, you should restart LiteFS. If you're using the Fly Platform, close
 your ssh session, and then restart your app:
 
 ```sh
@@ -104,7 +104,7 @@ elsewhere), you may see the following errors in your logs:
 
 You have two options to resolve this:
 
-- change the Consul key value that LiteFS uses. This is easy, but somewhat hacky/ugly.
+- change the Consul key value that LiteFS uses. This is easy, but leaves extra Consul keys set.
 - remove the "wrong" key from Consul. This is more difficult but cleaner.
 
 ### Easy option: change the Consul key
@@ -135,6 +135,7 @@ or just connect to your primary node via ssh and install it there. Here's how to
 ```sh
 # For debian/ubuntu-based images
 apt-get install consul
+
 # For alpine-based images
 apk add consul
 ```
@@ -148,7 +149,7 @@ echo $FLY_CONSUL_URL
 This url is in the format `https://:{TOKEN}@{HOST}/{PREFIX}`. You'll need each of these values
 separately. You'll also need the value of `lease.consul.key` from your `litefs.yml` file.
 
-Then you can delete the key with:
+Then you can use the [Consul CLI][] to delete the key with:
 
 ```sh
 export TOKEN=(copy the token you found from your consul url)
@@ -161,6 +162,8 @@ consul kv delete -http-addr=https://$HOST -token=$TOKEN $PREFIX/$LITEFS_CONSUL_K
 ```
 
 After this is done, restart your app.
+
+[Consul CLI]: https://developer.hashicorp.com/consul/commands/kv
 
 ## Add your replicas back
 
--- a/litefs/disaster-recovery.html.md.erb
+++ b/litefs/disaster-recovery.html.md.erb
@@ -1,0 +1,174 @@
+---
+title: Disaster Recovery from LiteFS Cloud
+layout: docs
+sitemap: false
+nav: litefs
+---
+
+The biggest advantage of using LiteFS Cloud is that you have backups
+to restore from in case of an unfortunate situation where you lose
+all the nodes in your cluster. In this doc, we'll show you how to restore
+your cluster from LiteFS Cloud if you've lost all the nodes.
+
+<div class="callout">
+You should only follow the steps in this document if you've lost all
+nodes of your LiteFS cluster. Otherwise, you can
+[restore from a LiteFS Cloud backup](/docs/litefs/cloud-restore).
+</div>
+
+## Start your app up with only the primary node
+
+It's simpler to recover a single node, rather than several nodes, so to
+start out with, it's best to remove all replica nodes and run just a
+primary node. If your app has already restarted, you may need to temporarily
+remove some replicas. If you're running your app on the Fly Platform, you can
+see the machines running your app with:
+
+```sh
+fly status
+```
+
+You'll see something like this:
+
+```
+Machines
+PROCESS	ID            	VERSION	REGION	STATE  	ROLE   	CHECKS	LAST UPDATED
+app    	784e900c2de378	1      	ams   	started	primary	      	2023-07-10T13:05:40Z
+app    	e784e471b39208	1      	ord   	started	replica	      	2023-07-10T13:01:16Z
+```
+
+Then destroy each machine that's marked `replica`:
+
+```sh
+fly machines destroy e784e471b39208
+```
+
+You can re-run `fly status` after you're done, to confirm you only have the primary running.
+
+## Find the correct cluster id
+
+When your app restarted, LiteFS generated a new cluster id, which is different
+than the cluster id stored in LiteFS Cloud. This is a safety measure: we don't want
+to accidentally overwrite data in a cluster if you inadvertently set the wrong LiteFS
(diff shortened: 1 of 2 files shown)


litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+```
+
+Then update the `clusterid` file:
+
+```sh
+echo LFSC25AD000F32ED02A9 > /var/lib/litefs/clusterid
+```
+
+After this, you should restart litefs. If you're using the Fly Platform, close
+your ssh session, and then restart your app:
```
benbjohnson, line 86: s/litefs/Litefs/

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+elsewhere), you may see the following errors in your logs:
+
+```
+2023-07-10T13:30:25Z app[784e900c2de378] ams [info]level=INFO msg="cannot connect, \"consul\" lease already initialized with different ID: LFSCA5E3680C1C2FCFFE"
+```
+
+You have two options to resolve this:
+
+- change the Consul key value that LiteFS uses. This is easy, but somewhat hacky/ugly.
+- remove the "wrong" key from Consul. This is more difficult but cleaner.
```
benbjohnson, line 108: I worry about using the phrase "hacky/ugly". It makes it sound like a bad option but the only real tradeoff is that you have extra Consul keys. Can we rephrase this to simply state the tradeoff rather than say one is better than another?

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+First, you'll need to add `consul` in your image. You can do this in your `Dockerfile` and re-deploy,
+or just connect to your primary node via ssh and install it there. Here's how to install it:
+
+```sh
+# For debian/ubuntu-based images
+apt-get install consul
+# For alpine-based images
+apk add consul
+```
```
benbjohnson, line 140: Can you either make this two separate code blocks or just put a newline in the middle?

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+```sh
+export TOKEN=(copy the token you found from your consul url)
+export HOST=(copy the host you found from your consul url)
+export PREFIX=(copy the prefix you found from your consul url)
+export LITEFS_CONSUL_KEY=(copy the key from your litefs.yml file)
+
+# and then delete the key!
+consul kv delete -http-addr=https://$HOST -token=$TOKEN $PREFIX/$LITEFS_CONSUL_KEY/clusterid
+```
```
benbjohnson, line 161: It might be worth mentioning that they can use `consul kv export` to list out their keys so they can find the one they're looking for.

```
consul kv export -http-addr=https://$HOST -token=$TOKEN
```

benbjohnson (approved):
Looks good! I just added a few minor comments & suggestions.

litefs/disaster-recovery.html.md.erb
magdalene: I tried to add this in, but it got confusing. I _think_ it's better to leave it as simple/straight path as possible. As a compromise though I can link [the consul kv CLI docs](https://developer.hashicorp.com/consul/commands/kv).

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+Note the cluster id specified at the end of the log message: `LFSC25AD000F32ED02A9` in
+our example. Copy this value to use later.
+
+## Update the cluster id
+
+Now that you have the correct cluster id, you can connect to your primary node and
+update the cluster id file. If your app is deployed on the Fly Platform, you can
+connect with:
```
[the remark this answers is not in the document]
magdalene: Andrea suggested "Fly Platform" [here](https://github.com/superfly/docs/pull/836/files#r1246748499), but I may be over-using it at this point.

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+export TOKEN=(copy the token you found from your consul url)
+export HOST=(copy the host you found from your consul url)
+export PREFIX=(copy the prefix you found from your consul url)
+export LITEFS_CONSUL_KEY=(copy the key from your litefs.yml file)
+
+# and then delete the key!
+consul kv delete -http-addr=https://$HOST -token=$TOKEN $PREFIX/$LITEFS_CONSUL_KEY/clusterid
+```
+
+After this is done, restart your app.
```
andie787, line 163: suggestion: include a command like you do elsewhere, is it `fly deploy` or `fly app restart`?

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+## Add your replicas back
+
+Once your primary node is connected to LiteFS Cloud again and your data has
+been recovered, you should go ahead and scale back up. If you're using the Fly
```
andie787, line 168: suggestion: start a new paragraph with "If you're using the Fly Platform"

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+or just connect to your primary node via ssh and install it there. Here's how to install it:
+
+```sh
+# For debian/ubuntu-based images
+apt-get install consul
+# For alpine-based images
+apk add consul
+```
+
+Then, take a look at your `FLY_CONSUL_URL`:
```
andie787, line 142: suggestion: "Get your `FLY_CONSUL_URL`:

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+
+You have two options to resolve this:
+
+- change the Consul key value that LiteFS uses. This is easy, but somewhat hacky/ugly.
+- remove the "wrong" key from Consul. This is more difficult but cleaner.
+
+### Easy option: change the Consul key
+
+You can simply update the `lease.consul.key` value in the `litefs.yml` file. The old one will
+still exist in Consul, but LiteFS won't care. Here's how to update your `litefs.yml`:
```
andie787, line 113: question: is this what the change needs to be? or can they put whatever they want?

suggestion: For example, you can just add `-v2` to the end of the Consul key value:

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+If you're using the Fly Platform, redeploy your app:
+
+```sh
+fly deploy
+```
+
+### Cleaner option: remove the wrong key from Consul
+
+First, you'll need to add `consul` in your image. You can do this in your `Dockerfile` and re-deploy,
+or just connect to your primary node via ssh and install it there. Here's how to install it:
```
andie787, line 133: question: how do they add consul in the Dockerfile? If we prefer the ssh install method, then maybe just have a "You can also do this in your Dockerfile..." note?

andie787 (commented):
Nice! I wonder if it might be a good idea to use numbered steps in the sections with more than 1 or 2 steps?

litefs/disaster-recovery.html.md.erb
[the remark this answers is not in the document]
andie787: yeah this came from my last edits! We're trying to use "the Fly Platform" to encompass Fly Apps and Fly Machines, the thing people can build anything on. It replaces "Fly Apps V2 platform" and is the one way we use Platform now. (I guess we need to write about this ;) )

litefs/disaster-recovery.html.md.erb
```diff
@@ -0,0 +1,174 @@
...
+sitemap: false
+nav: litefs
+---
+
+The biggest advantage of using LiteFS Cloud is that you have backups
+to restore from in case of an unfortunate situation where you lose
+all the nodes in your cluster. In this doc, we'll show you how to restore
+your cluster from LiteFS Cloud if you've lost all the nodes.
+
+<div class="callout">
```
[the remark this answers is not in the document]
magdalene: I thought about this and originally decided to have it as info rather than warning, because they'll still be able to recover from whichever LiteFS Cloud snapshot they want. But it will be a big hassle, and deleting data is always bad.... so I'll switch to warning!

litefs/disaster-recovery.html.md.erb
magdalene: It can be whatever they want. I updated with your wording, thanks!

litefs/disaster-recovery.html.md.erb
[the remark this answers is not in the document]
andie787: @catflydotio Yes, let's discuss. I am not attached to any particular wording. :)

PR #883, opened 2023-07-12
Trying out a content pattern and numbered steps in LiteFS doc

andie787:
Here's a partial preview of what the pattern might look like for LiteFS docs that work for non-Fly users and then also include specific steps for Fly Apps:

Screenshot 2023-07-12 at 8 02 30 PM


andie787:
@magdalene Closing this and we can revisit when more content is added to the docs so we can see what sort of template or pattern is needed.

PR #890, opened 2023-07-17
Improve machines API navigation

benwaffle:
| Old | New |
|-------|--------|
| image | image |


andie787 (commented):
This is a good thing! Suggest moving "Delete a Fly App" under "Operations on Apps". 

Minor thing: a lot of markdown style guides say not to have a heading with no text under it. But I don't think we have a rule about that yet. ;)

PR #898, opened 2023-07-23
Remove extra `9` from duration

tycol7:
Otherwise the command fails with 
```
Error: invalid argument "9999999h" for "-x, --expiry" flag: time: invalid duration "9999999h"
```


andie787:
Nice catch! Thanks!

PR #899, opened 2023-07-23
Add hint about specifying dockerfile in case it is not automatically …

tomczakdominik:
…detected

I have a Rails app which has a dockerfile named `Dockerfile` in a main directory and still `fly launch` was trying to build it as a ruby app, hence I propose adding this hint.

### DIFF
--- a/languages-and-frameworks/dockerfile.html.md.erb
+++ b/languages-and-frameworks/dockerfile.html.md.erb
@@ -31,7 +31,7 @@ Wrote config file fly.toml
 ? Would you like to deploy now?
 ```
 
-Let `fly launch` generate an app name for you or pick your own. You may need to explicitily specify Dockerfile path using `--dockerile` flag if fly doesn't detect Dockerfile automatically.
+Let `fly launch` generate an app name for you or pick your own. You may need to explicitly specify Dockerfile path using `--dockerfile` flag if fly doesn't detect Dockerfile automatically.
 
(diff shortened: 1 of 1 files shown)


languages-and-frameworks/dockerfile.html.md.erb
```diff
@@ -31,7 +31,7 @@ Wrote config file fly.toml
 ? Would you like to deploy now?
 ```
 
-Let `fly launch` generate an app name for you or pick your own.
+Let `fly launch` generate an app name for you or pick your own. You may need to explicitily specify Dockerfile path using `--dockerile` flag if fly doesn't detect Dockerfile automatically.
```
rubys, line 34: There are two typos in that one sentence.

rubys, line 34: The issue isn't that fly doesn't detect the Dockerfile automatically, successfully launching a Rails application typically requires the setting of at least one secret (RAILS_MASTER_KEY), and either the deployment of a database or the creation of a volume.  That's why the scanner for Rails is run before the Dockerfile scanner.

rubys (changes requested):

kcmartin:
Thanks for the suggestion, and sorry for the long delay! This page has since been rewritten, and the `--dockerfile` flag is now covered in the flyctl command reference. Closing this out, but thanks for flagging it.

PR #902, opened 2023-07-24
fly.toml ref: services.ports is required for a complete service config

catflydotio:


catflydotio:
I requested a few reviewers. You don't all have to review it!

reference/configuration.html.md.erb
```diff
@@ -291,15 +291,18 @@ Times are in milliseconds unless units are specified.
 
 ## The `services` sections
 
-The `services` sections configure the mapping of ports on the application to ports and services on Fly.io. These mappings determine how connections to the application will be handled on their journey from the Fly.io edge network to running Fly Apps.
+The `services` sections define how Fly Proxy connects requests to services running within Machines, via an app's public [Anycast](/docs/reference/services/) or private [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address. They also configure additional settings tied to Fly Proxy behavior for a service. If a service only needs HTTP and HTTPS on standard ports, you can use the less verbose [`http_service`](#the-http_service-section).
 
-You can have:
+A `services` section itself is a table of tables in TOML, so it is delimited with double square brackets. Tables within `[[services]]` configure additional features and behaviors for the service, and are described in subsequent sections. **At least one [`services.ports`](#services-ports) entry is required for a complete service definition, to set a port and handler for incoming requests.**
```
andie787, line 296: I think we either leave it plain text or put it in a note. We should avoid using bold in this way.

reference/configuration.html.md.erb
```diff
@@ -291,15 +291,18 @@ Times are in milliseconds unless units are specified.
 
 ## The `services` sections
 
-The `services` sections configure the mapping of ports on the application to ports and services on Fly.io. These mappings determine how connections to the application will be handled on their journey from the Fly.io edge network to running Fly Apps.
+The `services` sections define how Fly Proxy connects requests to services running within Machines, via an app's public [Anycast](/docs/reference/services/) or private [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address. They also configure additional settings tied to Fly Proxy behavior for a service. If a service only needs HTTP and HTTPS on standard ports, you can use the less verbose [`http_service`](#the-http_service-section).
```
andie787, line 294: I like that we bring Fly Proxy in here!
What does "via" mean in this case?
I'm not sure if the second sentence is needed. The "they" and the meaning of "Fly Proxy behaviour" isn't super clear.

reference/configuration.html.md.erb
```diff
@@ -291,15 +291,18 @@ Times are in milliseconds unless units are specified.
...
+
+* No `services` section (and no [`http_service` section](#the-http_service-section)): The application cannot be reached from the public internet, nor by a [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address; this is typically apps like databases that talk over [6PN private networking](/docs/reference/private-networking/) to other apps.
+* One `services` section (or an [`http_service` section](#the-http_service-section)): One internal port mapped to one or more external ports.
 * Multiple `services` sections (or an [`http_service` section](#the-http_service-section) and one or more `services` sections): Mapping multiple internal ports to multiple external ports.
+  
 
-The `services` section itself is a table of tables in TOML, so the section is delimited with double square brackets, like this:
+Example (note the double square brackets):
 
 ```toml
```
andie787, line 307: I wonder if the initial services example should include all the required sections? Or if we should have one complete example at the top and then break each piece out into the subsequent sections? And be more explicit about what's required and what's optional.

andie787: this might be Out Of Scope 👾

reference/configuration.html.md.erb
```diff
@@ -291,15 +291,18 @@ Times are in milliseconds unless units are specified.
...
 
-* No `services` section (and no [`http_service` section](#the-http_service-section)): The application has no mappings to the external internet - typically apps like databases that talk over [6PN private networking](/docs/reference/private-networking/) to other apps.
-* One `services` section (or an [`http_service` section](#the-http_service-section)): One internal port mapped to the one or more external ports on the internet.
+An app can have:
+
+* No `services` section (and no [`http_service` section](#the-http_service-section)): The application cannot be reached from the public internet, nor by a [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address; this is typically apps like databases that talk over [6PN private networking](/docs/reference/private-networking/) to other apps.
```
DAlperin, line 300: The line about databases could maybe be further qualified. I only say that since fly pg by default uses flycast

reference/configuration.html.md.erb
catflydotio: Oops! A whole coloured callout seemed like too big a deal and plain text too little, so I went with bold. Which do you think we should use?

(Is that a general convention, not to use bold inline?)

Adding it in plain text and moving the ports section up next to the top services section is more kinds of guidance than we had before. So plain text?

reference/configuration.html.md.erb
catflydotio:
```suggestion
The `services` sections define how Fly Proxy connects requests that hit an app's public [Anycast](/docs/reference/services/) or private [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address to services running within Machines. These sections also configure additional settings tied to Fly Proxy behavior for a service. If a service only needs HTTP and HTTPS on standard ports, you can use the less verbose [`http_service`](#the-http_service-section).
```

Better?

reference/configuration.html.md.erb
catflydotio: For this PR, I think so. 😁

reference/configuration.html.md.erb
catflydotio: Ah yes. That's something that was not true when this was written. I'll just cut it out.

reference/configuration.html.md.erb
catflydotio:
```suggestion
A `services` section itself is a table of tables in TOML, so it is delimited with double square brackets. Tables within `[[services]]` configure additional features and behaviors for the service, and are described in subsequent sections. At least one [`services.ports`](#services-ports) entry is required for a complete service definition, to set a port and handler for incoming requests.
```

reference/configuration.html.md.erb
```diff
@@ -291,15 +291,18 @@ Times are in milliseconds unless units are specified.
...
```
catflydotio, line 300:
```suggestion
* No `services` section (and no [`http_service` section](#the-http_service-section)): The application cannot be reached from the public internet, nor by a [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address; this is typical of apps that talk only over [6PN private networking](/docs/reference/private-networking/) to other apps on the same network.
```

reference/configuration.html.md.erb
```diff
@@ -291,15 +291,18 @@ Times are in milliseconds unless units are specified.
 
 ## The `services` sections
 
-The `services` sections configure the mapping of ports on the application to ports and services on Fly.io. These mappings determine how connections to the application will be handled on their journey from the Fly.io edge network to running Fly Apps.
+The `services` sections define how Fly Proxy connects requests that hit an app's public [Anycast](/docs/reference/services/) or private [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address to services running within Machines, along with settings for other Fly Proxy behavior for a service. If a service only needs HTTP and HTTPS on standard ports, you can use the less verbose [`http_service`](#the-http_service-section).
```
catflydotio, line 294:
```suggestion
The `services` sections define how Fly Proxy connects requests that hit an app's public [Anycast](/docs/reference/services/) or private [Flycast](/docs/reference/private-networking/#flycast-private-load-balancing) address to services running within Machines, and configure settings for other Fly Proxy behavior for a service. If a service only needs HTTP and HTTPS on standard ports, you can use the less verbose [`http_service`](#the-http_service-section).
```

catflydotio:
Calling it good to merge; always room to improve.

PR #905, opened 2023-07-26
Update fly.toml docs with missing config options

andie787:
This PR adds some missing config options for `fly.toml`, like `proxy_proto_options`. and `default_self_signed` for `tls_options`. Also updated and slighlty re-arranged the connection handler info in the Public Network Services doc. 

Related: 
- https://github.com/superfly/flyctl/pull/2169
- https://github.com/superfly/flyctl/pull/2177

### DIFF
--- a/reference/configuration.html.md.erb
+++ b/reference/configuration.html.md.erb
@@ -262,6 +262,18 @@ In the following example, `Example-Header` will be removed from the final respon
     Example-Multi-Value = ["value1", "value2"]
 ```
 
+### `http_service.proxy_proto_options`
+
+Configure the version of the PROXY protocol that your app accepts. Version 1 is the default.
+
+For example:
+```toml
+[http_service.proxy_proto_options]
+  version = "v2"
+```
+
+* `version` : A string to indicate that the TCP connection uses PROXY protocol version 2. The default when not set is version 1.
+
 ### `http_service.tls_options`
 
 Configure the TLS versions and ALPN protocols that Fly's edge will use to terminate TLS for your application with:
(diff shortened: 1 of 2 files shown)


reference/configuration.html.md.erb
```diff
@@ -262,6 +262,44 @@ In the following example, `Example-Header` will be removed from the final respon
     Example-Multi-Value = ["value1", "value2"]
 ```
 
+### `http_service.proxy_proto_options`
+
+Configure the version of the PROXY protocol that your app accepts. Version 1 is the default.
```
dangra, line 267: The confusing part is that for proxy_protocol to be used, the port handler must be `proxy_proto`

```toml
[[services]]
internal_port = 8080
protocol = "tcp"

[[services.ports]]
handlers = ["proxy_proto"]
proxy_proto_options = { version = "v2" } 
```

btw, now that I think about it, there is no way to set `proxy_proto` port handler on `http_service`, only for the general `services` section

dangra: I think we have to remove the proxy_proto_options from http_service, it only makes sense in the `services` section where you can control the ports and handlers.

dangra: I removed and added tests so we are sure it works for `services`  too https://github.com/superfly/flyctl/pull/2630

reference/configuration.html.md.erb
```diff
@@ -262,6 +262,44 @@ In the following example, `Example-Header` will be removed from the final respon
     Example-Multi-Value = ["value1", "value2"]
 ```
 
+### `http_service.proxy_proto_options`
```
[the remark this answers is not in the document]
dangra: nope, removing

PR #917, opened 2023-08-04
Clairfy required base64 encoding of file content references

neilmock:
The docs are misleading on what values need to be `base64` encoded, particularly w/r/t secret values that hydrate files, so hopefully this clarifies.

### DIFF
--- a/reference/configuration.html.markerb
+++ b/reference/configuration.html.markerb
@@ -674,7 +674,7 @@ This feature should not be compared directly with a CDN, for the following reaso
 
 ## The `files` section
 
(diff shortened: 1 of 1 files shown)


andie787:
@neilmock Would you mind clarifying what's misleading? I'm not sure of the difference between
- The contents of ... `secret_name` must be base64 encoded.
and your suggestion
- "...the value of the secret must be base64 encoded."

The content of `secret_name` is the value of the secret right?

kcmartin:
Thanks for this, and sorry for the long delay! The `files` section has since been rewritten and now spells out that it's the secret's value that must be base64 encoded, with an example. Since it's covered now, I'll close this out. Appreciate the flag!

PR #820, opened 2023-06-12
Update Fly Log Shipper Queue documentation

francoabaroa:
Update documentation to show users how to avoid duplicate logs in High Availability apps to avoid confusion

Sources:

https://community.fly.io/t/fly-log-shipper-duplicate-logs-on-apps-v2/13505

https://github.com/superfly/fly-log-shipper

PR #930, opened 2023-08-16
Update log shipper docs to ensure that no public access is granted

jsierles:
A customer [pointed out](https://community.fly.io/t/fly-log-shipper-publishes-graphql-publicly/14836/2) that our log shipper could expose its GraphQL API. We want standard shipper deployments to have no services nor IPs.

Note: `--no-public-ips` fails when an HTTP service is setup by default by `fly launch`.

### DIFF
--- a/going-to-production/monitoring/exporting-logs.html.md
+++ b/going-to-production/monitoring/exporting-logs.html.md
@@ -32,7 +32,7 @@ mkdir logshippper
 cd logshippper
 
 # Create the app but don't deploy just yet
-fly launch --no-deploy --image ghcr.io/superfly/fly-log-shipper:latest
+fly launch --no-deploy --no-public-ips --image ghcr.io/superfly/fly-log-shipper:latest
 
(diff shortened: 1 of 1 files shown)


andie787:
Should we mention that if they want only 1 machines, then use the `--ha=false` flag on `fly deploy`?

Related: this user PR for apps that do have 2 machines https://github.com/superfly/docs/pull/820

jsierles:
Yes, actually this may be a related issue. The shipper can operate in HA mode by setting a `QUEUE` env var, but that doesn't happen by default. I've been working on automating the shipper, but parked that for now.

andie787:
@jsierles Should we still ship this docs change or did it get fixed in flyctl?

kcmartin:
Thanks for this, and sorry it sat so long! The file moved since, so I've re-applied the fix (remove the public services section, deploy with `--no-public-ips`, plus notes for releasing existing IPs) in #2420. Closing in favor of that.

PR #940, opened 2023-08-23
[python] Update Python Docs: Flask App

katiayn:


languages-and-frameworks/python.html.markerb
```diff
@@ -7,198 +7,278 @@ redirect_from: /docs/getting-started/python/
...
 
-One thing to know about the builtin Python builder is that it will automatically copy over the contents of the  directory to the deployable image. This is how you can move static assets such as templates and other files to your application. The other thing to know is that it uses a Procfile to run the application; Procfiles are used on other platforms to deploy Python applications so we keep it simple. The Procfile contains instructions for starting the application. Here's the contents of ours:
+One thing to know about the built-in Python builder is that it will automatically copy over the contents of the directory to the deployable image. This is how you can move static assets such as templates and other files to your application.
+
+The other thing to know is that it uses a `Procfile` to run the application. `Procfiles` are used on other platforms to deploy Python applications so we keep it simple. The Procfile contains instructions for starting the application. Here's the contents of ours:
```
Roadmaster, line 180: The auto-created Procfile has this, which is different from what you show:
```
$ cat Procfile 
# Modify this Procfile to fit your needs
web: gunicorn server:app
```

Is it worth mentioning that one has to edit the Procfile to point to `app:app` instead?

Roadmaster (approved):
+1 I followed the instructions to the letter and they work with *one* small nitpick, see inline comment.

languages-and-frameworks/python.html.markerb
katiayn: @Roadmaster thanks for the suggestion! I'll update this part and mention to update the Procfile in that case. 🙂

katiayn:
@Roadmaster I mentioned in the docs to modify the Procfile if needed. I think it's better to update flyctl [here](https://github.com/superfly/flyctl/blob/master/scanner/templates/python/Procfile#L2) to reflect on more common app names like app.py or hello.py.

I used app.py so we don't need to set FLASK_APP environment variable, run FLASK_APP=hello flask run neither run flask --app hello run. Seems easier for beginners.

> The name “app” or “wsgi” is imported (as a “.py” file, or package), automatically detecting an app (app or application) or factory (create_app or make_app). ([reference](https://flask.palletsprojects.com/en/latest/cli/))

Having app.py, it's automatically detected as a Flask app.

WDYT?

Roadmaster:
Yes, having flyctl use app by default sounds better than using server (which I've never seen used anywhere!) and it will be easy to implement as it's just changing the Procfile template.

Down the line, how about checking for app and hello as Python modules and use either in the Procfile with a message that we guessed it was that module but to review the Procfile, and if they're not present, configure an obviously bogus module <YOUR_MODULE>:app and spit out a message that we couldn't find anything plausible and to edit the Procfile or things won't work.

It should be OK to run the Python interpreter to do a bit of module spelunking and importing to see what they offer (e.g. if there's an `app` module with an `app` attribute it's a relatively safe bet that's the one we want).

I think this should cover the most common frameworks outside of Django, that is, Flask, FastAPI and maybe Quart. Not sure if it's worth doing a ton of extra work for more esoteric ones (Pylons? Turbogears? Zope?) - but to be fair I have not seen these in the wild in years (with the exception of Zope, heh, but everyone in that org recognized we were total outliers using that).

Roadmaster (approved):
+1 as previously :)

katiayn:
@Roadmaster I did some research and checked some of the most used/popular python frameworks:
- Flask
- FastAPI
- Pyramid
- Tornado
- CherryPy
- Bottle
- Web2py
- Kivy

Well, there are a lot of them! 😁 I updated the docs for "Run a Flask App" and I also prepared a FastAPI app which I'll add next - that's why I changed the layout a bit.

Screenshot 2023-08-25 at 17 31 06

I'll add "Run a FastAPI App" next and keep some of the common docs for both. The idea is to have other frameworks added later on. But for now, those are main frameworks.

Since Gunicorn by itself is not compatible with FastAPI, the `Procfile` will need to be updated for that.

To simplify, I'll just update the template in `flyctl` to `app:app` and **highlight in the docs** how to modify the `Procfile` according to the framework - there is no default `Procfile` for Flask and FastAPI apps. It seems better because there are a LOT of use cases and different ways to implement Flask/FastAPI apps.

Roadmaster:
@katiayn +1 I agree on the approach, focusing on the most popular non-Django frameworks first (Flask is definitely up there, I'd venture to say it's more popular than Django by now and FastAPI is the new thing!) makes sense. Sure, there's no common approach since Flask uses gunicorn mainly and FastAPI relies on uvicorn. As for the others, how about mentioning in an "other frameworks" section that all you need to do is update the Procfile with the command typically used to start your web server in production mode pointed at the correct module or something similar?

Also don't forget about [Quart](https://pgjones.gitlab.io/quart/)! (async Flask basically).

Thanks for fleshing this out, it's looking great :)

(Also Github continues not to alert me to your comments :( I don't know why. I'll try watching the MP to see if I get more notifications that way).

katiayn:
@Roadmaster I'm merging this PR. I'm working on the FastAPI docs - later on I'll add more specific info about some of the other frameworks. 

I'm rethinking how to approach these docs for multiple frameworks - maybe something like what Sam did for JS [here](https://fly.io/docs/js/).

Also thinking about removing most of the explanation to have a much simpler "Getting Started" docs. I'm curious about what you think 😁

PR #962, opened 2023-09-04
Adapt to deprecations in superfly/flyctl#2733

catflydotio:
This is a quick adaptation to recent flyctl changes (superfly/flyctl#2733). Replaces deprecated top-level commands with their `fly apps` equivalents in docs, and removes deprecated commands from the flyctl nav.

### DIFF
--- a/apps/scale-count.html.markerb
+++ b/apps/scale-count.html.markerb
@@ -324,4 +324,4 @@ If a Machine is misbehaving (for instance, it's not `stop`ping successfully), yo
 fly machine destroy --force 0e286039f42e86
 ```
 
-<div class="callout"> If you `destroy` a Machine, any volume that it had been using still exists until you either `fly volumes delete` it or `fly destroy` the Fly App it belongs to.</div>
+<div class="callout"> If you `destroy` a Machine, any volume that it had been using still exists until you either `fly volumes delete` it or `fly apps destroy` the Fly App it belongs to.</div>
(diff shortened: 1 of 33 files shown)


apps/scale-count.html.markerb
```diff
@@ -324,4 +324,4 @@ If a Machine is misbehaving (for instance, it's not `stop`ping successfully), yo
 fly machine destroy --force 0e286039f42e86
 ```
 
-<div class="callout"> If you `destroy` a Machine, any volume that it had been using still exists until you either `fly volumes delete` it or `fly destroy` the Fly App it belongs to.</div>
+<div class="callout"> If you `destroy` a Machine, any volume that it had been using still exists until you either `fly volumes delete` it or `fly apps destroy` the Fly App it belongs to.</div>
```
andie787, line 327: My preference would be not to use commands as verbs like this. In this case, remove the backticks from "If you destroy a Machine" and rewrite the rest like "until you either delete the volume with the `fly volumes delete` command or delete the app the volume belongs to with the the `fly apps destroy` command."

andie787 (commented):
That's a lot of `fly open`s! Left one comment.

apps/scale-count.html.markerb
catflydotio: Oh no!

catflydotio, line 327:
```suggestion
<div class="callout"> If you destroy a Machine, any volume that it had been using remains intact until you either explicitly destroy the volume or destroy the Fly App it belongs to.</div>
```

catflydotio: I tidied the callout. My reasoning here is that it's unnecessarily long to include all the commands.

PR #965, opened 2023-09-05
Remove TCP idle timeout restrictions

kylemclaren:
Per: https://community.fly.io/t/tcp-idle-timeouts-restrictions-have-been-removed/15160

I have removed the whole paragraph, but open to adding _something_ in its place


catflydotio:
Actionable info about the new behaviour would be great but clearing out now-misleading info is a worthwhile start.

PR #972, opened 2023-09-08
Working with support

kylemclaren:


Roadmaster:
This is great :) Would it be good to add these (perhaps obvious) expectations as explicit?

Tickets should include:
1. What you were trying to achieve (e.g. deploy the application)
2. What you did (e.g. I ran fly deploy)
3. What you observed ('it threw a 500 error')

(I'm eliding the traditional "what you expected to happen" because we can usually infer that from #1, and it avoids the common "I expected it to freaking work!" reaction)
:boom:

For "the app stopped responding" or "it's slow" kind of tickets, even #1 can lead to that ("I was trying to achieve 100% uptime with 100ms response times!). But still, a description of what they were trying or wanted the end result to be, would be good as a starting point.

about/support.html.md
```diff
@@ -30,5 +30,16 @@ Customers on [Launch, Scale, and Enterprise plans](https://fly.io/plans) have ac
 
 Your support address is in the [Fly.io dashboard](https://fly.io/dashboard). Select your organization, and look for your Support email address in the Billing section.
 
+## Working with support
+We understand that describing software issues can sometimes be challenging. Providing a clear and detailed description of the problem, including any noticeable symptoms and relevant artifacts, can greatly assist us in identifying similar cases and expediting the resolution process. If we can quickly find a similar issue, we'll work to resolve it promptly. Conversely, if the problem is more complex, we'll involve specialized resources to address it.
```
Roadmaster, line 34: maybe "software and infrastructure issues" or "technical issues" since it's not always just software?

Roadmaster (approved):
+1 from me, I think it's good to make this explicit; not everybody will read this prior to filing a ticket but even if they don't, we'll have a consistent response to send when we get an incomplete ticket.

A few inline comments ...

PR #977, opened 2023-09-12
The utility is flyctl.

adelton:


andie787:
The utility is flyctl and you could run it as `flyctl auth signup`, but we prefer the shorter `fly auth signup`. You can run all flyctl commands with `fly <command>`.

Issue #978, opened 2023-09-12
Additional \ showing in the docs


redjonzaci:
If this is something I can fix, let me know!

<img width="471" alt="image" src="https://github.com/superfly/docs/assets/73707194/c34975e9-27e1-49cb-9c80-ee36fbc44973">

andie787:
A few of these seem to left over! If you want to open a docs PR, feel free. One of us can test it out.

redjonzaci:
I just figured out how to do it.

redjonzaci:
I have 2 more suggestions:
- renaming the page where I found the backslashes (https://fly.io/docs/elixir/getting-started/) to "Getting Started with Elixir" for better SEO.
- opening "Edit this page with GitHub" links in a new tab because the contributor might need to compare the page with the GitHub representation (I needed to do that).

If you decide to consider the suggestions, I can work on them, maybe even in this PR, or new PRs.

redjonzaci:
The first suggestion is not very necessary, but this is how it shows in search results:
<img width="480" alt="image" src="https://github.com/superfly/docs/assets/73707194/def0905d-d39b-4d93-9594-6f4a3299e1eb">

andie787:
Thanks for the suggestions; these are good points. We'll consider them as we continue to update the docs. I'm going to close this now though because you fixed the issue :)

PR #980, opened 2023-09-12
Escape backslashes

redjonzaci:
Related to: #978 

The `app_name` also has backslashes, but didn't know how to change it.

### DIFF
--- a/languages-and-frameworks/partials/_launch_with_postgres.html.erb
+++ b/languages-and-frameworks/partials/_launch_with_postgres.html.erb
@@ -3,7 +3,7 @@
 When we run `fly launch` from the newly-created project directory, the launcher will:
 
 * Ask you to select a deployment region
-* Set secrets required by <%= detected %> (`SECRET\_KEY\_BASE`, for example)
+* Set secrets required by <%= detected %> (`SECRET\\_KEY\\_BASE`, for example)
 * Run the <%= detected %> deployment setup task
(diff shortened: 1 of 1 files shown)


languages-and-frameworks/partials/_launch_with_postgres.html.erb
```diff
@@ -3,7 +3,7 @@
 When we run `fly launch` from the newly-created project directory, the launcher will:
 
 * Ask you to select a deployment region
-* Set secrets required by <%= detected %> (`SECRET\_KEY\_BASE`, for example)
+* Set secrets required by <%= detected %> (`SECRET\\_KEY\\_BASE`, for example)
```
andie787, line 6: Here and elsewhere, you can remove the backslashes, since they aren't needed to escape the `_` and `:` characters any more.

andie787 (changes requested):
A few other partials have this issue as well. Check for a ruby method that modifies the `app_name` variable. That bit of code can be removed in this PR.

languages-and-frameworks/partials/_launch_with_postgres.html.erb
redjonzaci: I removed only them in this file for now.

redjonzaci:
Sorry! I missed that.

andie787:
Looks good. Thanks for noticing it

PR #983, opened 2023-09-12
Update conflicting default value

redjonzaci:
I read in https://fly.io/docs/reference/configuration/#the-http_service-section 
that the default value for `auto_stop_machines` is `true` while this page conflicts that.

### DIFF
--- a/apps/autostart-stop.html.markerb
+++ b/apps/autostart-stop.html.markerb
@@ -41,7 +41,7 @@ Concurrency limits for services affect [how automatic starts and stops work](#ho
 
 ### Default values
 
-The default settings for apps that don't specify any auto start and stop settings in `fly.toml` are `auto_start_machines = true` and `auto_stop_machines = false`.
+The default settings for apps that don't specify any auto start and stop settings in `fly.toml` are `auto_start_machines = true` and `auto_stop_machines = true`.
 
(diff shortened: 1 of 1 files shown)


redjonzaci:
I also wonder if we could do some sort of sync for default and possible values based on the flyctl repo.

redjonzaci:
I think we don't even need that first sentence, since we show the default values below:

image

redjonzaci:
Should be `auto_stop_machines = true`?

image

redjonzaci:
Switch to "shutdown"?
https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/s/shut-down-shutdown

catflydotio:
> I think we don't even need that first sentence, since we show the default values below:
> 
> image

If you look carefully, these two paragraphs are talking about two different things. ;)

redjonzaci:
Maybe we should specify how the apps are created in the first paragraph?

redjonzaci:
Okay, did some more reading, and it makes sense to me now.
The only suggestions that are left:
- https://github.com/superfly/docs/pull/983#issuecomment-1716289743
- https://github.com/superfly/docs/pull/983#issuecomment-1716298124

andie787:
Please make new commits to update the content of the PR with the suggestions.

redjonzaci:
Also, if you need more help with the docs, let me know.

PR #996, opened 2023-09-14
Some ports were strings, I made them integers.

coldeny:
After asking here to make sure: https://community.fly.io/t/i-am-confused-on-why-port-is-an-integer-and-then-a-string-in-services-ports-example/15385


andie787:
Thanks for noticing!

PR #998, opened 2023-09-14
Minor improvements to "Private Networking" page

redjonzaci:
Also, the second column of the table in https://fly.io/docs/reference/private-networking/#fly-internal-addresses is named "aaaa". Is that intentional?


redjonzaci:
Included in #1030 #1031 #1032

PR #999, opened 2023-09-14
Remove duplicate "Volumes" entry from the navbar

redjonzaci:
We also have it under https://fly.io/docs/database-storage-guides/. (where it belongs?)


andie787:
This is by design so that users can find it under both reference and databases sections.

PR #1002, opened 2023-09-14
Format as code in "Volumes" page

redjonzaci:


andie787:
Request: It would make it easier for us if you could group all your tiny PRs into bigger PRs by page or type of review. Otherwise, we just won't have time to go through them all.

For example, you could group your suggestions for the Volumes page (this PR, #1001, #1002) into one PR
or
you could group your "format as code" PRs (this PR, #991, #992, #1003, #1004, #1010, #1012, #1017) into one PR.

This includes all of the "minor improvements" PRs as well if there's a theme going there.

redjonzaci:
Of course, the reason for that is that every time I read a page, I submitted the PR so as not to forget the improvements.

I will apply your suggestion to my current PRs.

Additionally, I will keep note of every improvement and submit PRs for groups of changes daily/weekly.

andie787:
I'm not sure if it's enabled for contributors, but you could try opening a draft PR that you can keep adding to until it's ready or until you've finished reviewing something. That way you can switch between branches and not lose anything.

Also, we'll have a linter soon, so the need is more for identifying gaps in docs and understanding the problems users face and how docs can help them, rather than copy editing. Is there a common problem in community that we don't have a doc for? Do users understand how to make the most of the features we have? That kind of thing.

redjonzaci:
I will join the current PRs in common groups and will then focus on higher-level issues in the docs.

redjonzaci:
I have a [Fly.io Ways of Working](https://redjonzaci.notion.site/redjonzaci/Fly-io-ca8639e2bbb14579b63b25f3abdb43c0#b8e7d84dc7854611aed016a3f4ab4508) section where I add the suggestions I get while contributing to `flyctl` and `docs` repos.

redjonzaci:
Included in #1030

PR #1006, opened 2023-09-14
Add link to Apps V2 page

redjonzaci:
- It's not obvious that the headings are links to other pages.


redjonzaci:
Included in #1031

PR #1007, opened 2023-09-14
Minor improvements to "This Is Not Managed Postgres" page

redjonzaci:
- TimescaleDB is included in the default image and can be enabled with these [instructions](https://fly.io/docs/postgres/managing/enabling-timescale/). <- the link is broken


redjonzaci:
Included in #1031 #1034

PR #1013, opened 2023-09-15
Improvements to "App Connection Examples" page

redjonzaci:


redjonzaci:
Also, the docs links below each example look a little weird. The ideal would be linking in the same line as the header, but also underlining it.

redjonzaci:
Included in #1034

PR #1024, opened 2023-09-15
No longer mention Hashicorp products as open-source

hkrutzer:
These products are now source-available instead of open-source.

The flow of the paragraph is changed by removing this sentence so perhaps it needs more work.


catflydotio:
Addressed by https://github.com/superfly/docs/pull/1040

PR #1029, opened 2023-09-16
Uppercase "http" and "tcp"

redjonzaci:


catflydotio:
I'd consider merging this, because it looks OK at first glance, but this kind of search-and-replace PR touches a lot of files and could create merge conflicts for higher-value docs work. (I see one merge conflict it would create already)

PR #1030, opened 2023-09-16
Format as code

redjonzaci:


catflydotio:
flyctl is a name and generally shouldn't be formatted as code.

redjonzaci:
I can remove that

PR #1031, opened 2023-09-16
Update links

redjonzaci:


redjonzaci:
In https://fly.io/docs/postgres/getting-started/what-you-should-know/:
TimescaleDB is included in the default image and can be enabled with these [instructions](https://fly.io/docs/postgres/managing/enabling-timescale/). <- the link is broken

redjonzaci:
In pages that have headings as links to other pages, it's not obvious that that those headings are links.

catflydotio:
It's a fair amount of work to read through a large number of very small, unrelated changes, and we're unlikely to merge such changes that touch a lot of files at once from a community PR. The occasional correction is great! But I'd encourage you to focus on changes or new information that make users' lives easier.

PR #1032, opened 2023-09-16
Update keyword casing

redjonzaci:


andie787 (changes requested):
1. Issues with capitalization in command line output should be left as-is. Source is flyctl repo.

2. Flyctl docs are generated automatically, so the source is in flyctl not docs repo. 

If you want, you can  open an issue in the flyctl repo listing the commonly misspelled words (for both command line output and flyctl docs). An issue might be better than PR, since the PR might get stale with so many files. You can flag me on the issue if you like.

redjonzaci:
Sure, I can leave out changes in `flyctl/cmd`.

redjonzaci:
And keeping note of the words.

andie787:
It's not that we don't want contributors to update them, it's just that an issue might be better in this case since you'd be updating over 20 source files and I'm not sure when it would get merged. You can feel free to open a PR if that's your preference!

redjonzaci:
I understand, but I meant contributors could submit the issue or PR to `flyctl` instead of this repo.
A single public source of truth so no need for us to explain this again.

But I guess not worth the work of separating them since this issue won't occur very often.

redjonzaci:
It should be ready to go now.

The misspelled words related to `flyctl` are:
- wireguard
- fly
- postgres

rails/cookbooks/databases.html.markerb
```diff
@@ -48,7 +48,7 @@ one:
         ```
 
 In general, requests that never make it to Rails are served fastest, so
-the second approach where fly serves the static files is the better approach.
+the second approach where Fly serves the static files is the better approach.
```
andie787, line 51:
```suggestion
the second approach where Fly.io serves the static files is the better approach.
```

rails/advanced-guides/terraform.html.md
```diff
@@ -56,8 +56,8 @@ bin/rails generate fly:terraform
 The terraform generator supports a number of options, for now it is worth letting them all default.  For completeness, the current list of options supported is:
 
   * `--name` - application name. If not specified, a name will be generated for you.
-  * `--org` - the fly organization. Defaults to `personal`.
-  * `--region` - one or more regions to deploy this app.  If not specified, fly will pick one for you based on your location.
+  * `--org` - the Fly organization. Defaults to `personal`.
+  * `--region` - one or more regions to deploy this app.  If not specified, Fly will pick one for you based on your location.
```
andie787, line 60:
```suggestion
  * `--region` - one or more regions to deploy this app.  If not specified, Fly.io will pick one for you based on your location.
```

rails/advanced-guides/machine.html.md
```diff
@@ -71,7 +71,7 @@ Now lets make that application launch more machines.
 
 ## Installing `fly` on the Rails Machine
 
-Since we will be using fly services from within our Rails application, we will need to install the fly executable.  We do that by adding the following lines to our `Dockerfile`:
+Since we will be using Fly services from within our Rails application, we will need to install the fly executable.  We do that by adding the following lines to our `Dockerfile`:
```
andie787, line 74:
```suggestion
Since we will be using Fly.io services from within our Rails application, we will need to install the fly executable.  We do that by adding the following lines to our `Dockerfile`:
```

andie787 (changes requested):

PR #1035, opened 2023-09-16
Update wording

redjonzaci:
Also, can we assume that developers know that CLI stands for "command line interface"
and update it in the command docs too?

### DIFF
--- a/rails/getting-started/index.html.md
+++ b/rails/getting-started/index.html.md
@@ -38,11 +38,21 @@ and deploy the application.
 
 ### Create an application
 
-Start by verifying that you have Rails version 7 installed, and then by
-creating a new application:
+You will need Ruby and Rails installed on your machine. If you do not have them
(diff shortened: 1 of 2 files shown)


andie787 (changes requested):

rails/getting-started/index.html.md
```diff
@@ -38,11 +38,17 @@ and deploy the application.
...
+
+Verify that Rails version 7 was installed correctly:
 
 ``` shell
 $ rails --version
+```
+
+Then create a new application and change into its directory:
+
+``` shell
```
andie787, line 51: Just want to make sure that @rubys is okay with these changes/suggestions?

rubys:
Rails docs don't use CLI, see: https://guides.rubyonrails.org/command_line.html

So, I'm not in favor of switching to CLI.  The other changes look fine.

PR #1042, opened 2023-09-18
Extract db connection page as a section in Laravel docs, remove larav…

KTanAug21:
…el dbs page.

This will remove the overcrowding in the laravel-database connection page, and create a new section where new, individual laravel-db connection pages can be added


laravel/database-guides/laravel-mysql.html.markerb
```diff
@@ -0,0 +1,123 @@
...
+    fly deploy 
+    ```
+
+
+
+### _Connect from a local environment_
+
+The MySQL instance you spun up in Fly.io "[is closed to the public internet](/docs/reference/machines/#notes-on-networking)", and can only be accessed by another application found in your Fly.io organization's private network. You'll need a way to tunnel into the network, and finally connect to your MySQL instance.
+
+<b>In this guide you'll tunnel to your MySQL instance through the use of `flyctl proxy`</b>
```
redjonzaci, line 46: Why not use fly instead of flyctl

redjonzaci:
I also found some typos that can be fixed, but maybe we don't want to focus on those since a linter is coming soon.

KTanAug21:
> I also found some typos that can be fixed, but maybe we don't want to focus on those since a linter is coming soon.

Hello @redjonzaci! Thanks for promptly checking this PR, and great catch on those typos! I'll add a new commit to lint them out

redjonzaci:
I may have used the wrong word to refer to inconsistencies in casing, wording and punctuation.
Basically, I saw some missing . in end of sentences, missing :, some text that needs to be formatted as code, incorrect casing of "fly app", "docker", etc.

PR #1043, opened 2023-09-19
Add machine restart policy doc

andie787:
### Summary of changes

- Added a doc to the Machines section about Machine restart policies
- Added some examples

### Related GitHub issues and community posts

Source info: https://community.fly.io/t/issues-with-machines-restart-policy/11943/25?u=andie

### Notes (optional)

n/a

### DIFF
--- a/machines/guides-examples/machine-restart-policy.html.markerb
+++ b/machines/guides-examples/machine-restart-policy.html.markerb
@@ -6,13 +6,15 @@ order: 35
 nav: firecracker
 ---
 
-A machine restart policy defines whether and how flyd restarts a Machine after its main process exits. The restart policy applies to an individual Machine, and is not an app-wide setting.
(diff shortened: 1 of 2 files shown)


machines/guides-examples/machine-restart-policy.html.markerb
```diff
@@ -0,0 +1,93 @@
...
+  	... // 6 identical lines
+
+? Apply changes? (y/N)
+```
+
+Enter 'y' to apply the changes.
+
+## Change a Machine's restart policy with the Machines API
+
+With the Machines API, you can set the restart policy, and the maximum number of retries when the policy is `on-fail`.
```
andie787, line 79: IS 'on-fail` and alias for `on-failure`. Might be better if the API and flyctl match?

jipperinbham: `on-failure` mimics Docker's naming which was the goal, if the CLI is using something different, we should change it.

jipperinbham (approved):

catflydotio (approved):

Issue #1067, opened 2023-09-23
Broken link on volumes documentation page


lpil:
Hello!

The anchor that reads "How to scale an app with volumes attached" on [this page](https://fly.io/docs/reference/volumes/) has an empty `href`, it doesn't go anywhere.

https://github.com/superfly/docs/blob/04ee4638e894c6045bbb0ca04fb14584f4ab2b24/reference/volumes.html.markerb#L47

Thanks,
Louis

Lucais11:
Thanks for bringing this to our attention, I'm on it!

PR #1069, opened 2023-09-24
Added the necessary url that points to the correct doc

Lucais11:
### Summary of changes
Added the necessary url to the href that points to the correct doc. Located on this page-`/docs/reference/volumes/`
#1067

### Notes
n/a if none

### DIFF
--- a/reference/volumes.html.markerb
+++ b/reference/volumes.html.markerb
@@ -44,7 +44,7 @@ Learn more about using volumes:
 
 - How to [add volume storage for your app](/docs/apps/volume-storage/), step-by-step.
 - How to configure the [`mounts` section](/docs/reference/configuration/#the-mounts-section) for volumes in the `fly.toml` Fly Launch configuration file.
-- How to [scale an app with volumes attached](https://fly.io/docs/apps/scale-count/#scale-an-app-with-volumes)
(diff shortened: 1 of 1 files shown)


reference/volumes.html.markerb
```diff
@@ -44,7 +44,7 @@ Learn more about using volumes:
 
 - How to [add volume storage for your app](/docs/apps/volume-storage/), step-by-step.
 - How to configure the [`mounts` section](/docs/reference/configuration/#the-mounts-section) for volumes in the `fly.toml` Fly Launch configuration file.
-- How to [scale an app with volumes attached]()
+- How to [scale an app with volumes attached](https://fly.io/docs/apps/scale-count/#scale-an-app-with-volumes)
```
andie787, line 47:
```suggestion
- How to [scale an app with volumes attached](/docs/apps/scale-count/#scale-an-app-with-volumes)
```
we use relative links in docs

andie787 (commented):
Thanks for doing this! Request a small change to use a relative link

reference/volumes.html.markerb
Lucais11: I didn't know if I should use the relative link or a specific link that points to where it talks about the words preceding the link, "Scale an app with volumes attached"

Issue #1065, opened 2023-09-22
GH Actions Deployment instructions don't work: generated token is not usable


anatoliykmetyuk:
The [Continuous Deployment](https://fly.io/docs/app-guides/continuous-deployment-with-github-actions/) page mentions generating a token via `fly tokens create deploy -x 999999h` for GitHub Actions. However, if I use that token, it doesn't work: I get the "Could not find App" error. What works is going to the [personal access tokens](https://fly.io/user/personal_access_tokens) page and generating a token from there, and using that token for GitHub Actions.

Lucais11:
Hey @anatoliykmetyuk, just to clarify, when you ran into the error `Could not find App` where you in the right directory when running the command- `fly tokens create deploy -x 999999h`? When you "used the token", where or how were you using it?

anatoliykmetyuk:
@Lucais11 I was in the directory of the app I was deploying, that is the same directory that contains `.github` and `fly.toml` files.

"Used the token" means used it as per instruction on the website: I went to the repo secrets, created a new secret and inserted there the token generated by the command. Note that when I used the same procedure with the token generated manually from dashboard, the deployment worked.

Also note that the token generated by `fly tokens create deploy -x 999999h` was MUCH longer than the one generated from dashboard. The dashboard one was around one line (80 characters or so) and the one from the command took 3 lines.

Lucais11:
@anatoliykmetyuk Thank you for clarifying all of that! 

There is a difference between tokens from fly.io > dashboard > account > access tokens and tokens from fly.io > dashboard > apps > tokens on the left sidebar > create deploy token. 

When running `fly tokens create deploy -x 999999h` you will see the `deploy token` in the app directory you ran the command, meaning you'll find it in fly.io > dashboard > app you ran the command in > tokens on the left sidebar > create deploy token. 

So I say all this to say, if you see your deploy token for the app you ran the command for, then it got created successfully when you ran `fly tokens create deploy -x 999999h`. Just to double check as well, you'd want to copy the entire output meaning even the part that starts with `FlyV1`. Insert the entire token into the Github secret.

anatoliykmetyuk:
> So I say all this to say, if you see your deploy token for the app you ran the command for, then it got created successfully when you ran fly tokens create deploy -x 999999h. Just to double check as well, you'd want to copy the entire output meaning even the part that starts with FlyV1. Insert the entire token into the Github secret.

[...]

PR #1071, opened 2023-09-26
Change the github action instructions, so that it doesn't show a spec…

Lucais11:
…ific access token(which was confusing customers)

### Summary of changes
Made a specific small change in the Continuous Deployment doc that shows an access token, which made customer's think they need to use that specific token, instead of generating one themselves. Link to doc- https://fly.io/docs/app-guides/continuous-deployment-with-github-actions/ 
#1065

### Notes
n/a if none

### DIFF
--- a/app-guides/continuous-deployment-with-github-actions.html.md
+++ b/app-guides/continuous-deployment-with-github-actions.html.md
@@ -24,7 +24,7 @@ The first section is a speed-run through the steps to make the go-example app au
 1.  Fork the [go-example](https://github.com/fly-apps/go-example) repository to your GitHub account.
 2.  Clone the new repository to your local machine.
 3.  Run `fly launch` from within the project source directory to create a new app and a `fly.toml` configuration file. Type `N` when `fly launch` asks if you want to set up databases and `N` when it asks if you want to deploy.
-4.  Still in the project source directory, get a Fly API deploy token by running `fly tokens create deploy -x 999999h`. Copy the output.
+4.  Still in the project source directory, get a Fly API deploy token by running `fly tokens create deploy -x “custom access token”`. The access token can be generated at https://fly.io/user/personal_access_tokens. Copy the output.
 5.  Go to your newly-created repository on GitHub and select **Settings**.
 6.  Under **Secrets and variables**, select **Actions**, and then create a new repository secret called `FLY_API_TOKEN` with the value of the token from step 4.
 7.  Back in your project source directory, create `.github/workflows/fly.yml` with these contents:


app-guides/continuous-deployment-with-github-actions.html.md
```diff
@@ -24,7 +24,7 @@ The first section is a speed-run through the steps to make the go-example app au
 1.  Fork the [go-example](https://github.com/fly-apps/go-example) repository to your GitHub account.
 2.  Clone the new repository to your local machine.
 3.  Run `fly launch` from within the project source directory to create a new app and a `fly.toml` configuration file. Type `N` when `fly launch` asks if you want to set up databases and `N` when it asks if you want to deploy.
-4.  Still in the project source directory, get a Fly API deploy token by running `fly tokens create deploy -x 999999h`. Copy the output.
+4.  Still in the project source directory, get a Fly API deploy token by running `fly tokens create deploy -x “custom access token”`. The access token can be generated at https://fly.io/user/personal_access_tokens. Copy the output.
```
catflydotio, line 27:
```suggestion
4.  Still in the project source directory, get a Fly API deploy token by running `fly tokens create deploy -x <your-access-token>`. The access token can be generated at https://fly.io/user/personal_access_tokens. Copy the output.
```

Our convention is to put angle brackets around "variables" in CLI examples.

This suggestion is based on the assumption that you don't actually need double quotes around the token string -- if you do, we need to add the double quotes back in.

anatoliykmetyuk: Oh, I thought `9999999h` was the lifetime of the token, i.e. "expire in that many hours".

But it's still not clear to me why run `fly tokens create deploy -x “custom access token”` with the custom token at all? For me, the deployment worked entirely without running that command, I just generated the token at the dashboard and inserted it directly into GH repo secrets. What is the purpose of this command?

andie787: hi @anatoliykmetyuk 
You're right that the value of `-x` is the token expiry!

Sorry, talked about this internally with @Lucais11 yesterday, but we didn't get back to this PR yet to fix it.

(Per the discussion over in #1065, we weren't sure if you got the error when you ran the `fly tokens create deploy` command or when you tried to use the token you got from that command. Since the error happened when you tried to use the token, then we'll have to look into what happened there.)

The short version is that you solved your issue by using a different kind of token.

andie787: I realized I didn't answer all your questions. Just to clarify, you would not run `fly tokens create deploy -x “custom access token”`, that was a misunderstanding. 
The command `fly tokens create deploy` creates a deploy token. A deploy token is scoped to the app you created it for, while the token you created in the UI under **Account** is a user-scoped OAuth token.

anatoliykmetyuk: > we weren't sure if you got the error when you ran the fly tokens create deploy command or when you tried to use the token you got from that command. Since the error happened when you tried to use the token, then we'll have to look into what happened there.) The short version is that you solved your issue by using a different kind of token

Yes, to clarify, `fly tokens create deploy` worked just fine, but the token it produced did not work when I was actually deploying from GH Actions.

andie787:
@Lucais11 I think we can close this now?

PR #1072, opened 2023-09-26
Update Github actions yml

SaraVieira:
### Summary of changes
Updates the yaml for the example github action to use `main` instead of `master` as that is the most common branch name now and the one recommended by github

https://fly.io/docs/app-guides/continuous-deployment-with-github-actions/

### Notes
n/a if none

### DIFF
--- a/app-guides/continuous-deployment-with-github-actions.html.md
+++ b/app-guides/continuous-deployment-with-github-actions.html.md
@@ -33,7 +33,7 @@ The first section is a speed-run through the steps to make the go-example app au
     on:
       push:
         branches:
-          - master
+          - main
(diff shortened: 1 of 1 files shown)


andie787:
hey @SaraVieira 
you're right that `main` is preferred over `master` for repos, but this tutorial uses an example repo where the branch is `master`. We added this note awhile ago after the yaml file example as a stop-gap until we can rename things:

> Note that the go-example’s default branch is currently master. If you’re using a different app, yours might be main. Change the fly.yml file accordingly.

SaraVieira:
Fair! Sorry i didn't see the repo this was connected to

PR #1074, opened 2023-09-27
Specify in fly ips release command that you have to pass the address …

Lucais11:
…of the ip or it will fail

### Summary of changes
Specified where `fly ips release` was mentioned to include `<address>` at the end of the command, because if no address is passed in, the command fails
n/a if none

### Notes
n/a if none


andie787 (approved):
LGTM! 🚀

PR #1078, opened 2023-09-29
Add host down troubleshooting doc

andie787:
### Summary of changes

- add a doc with steps to take to try to recover non-HA apps from hosts with issues

n/a

### Notes
n/a

### DIFF
--- a/apps/trouble-host-unavailable.html.markerb
+++ b/apps/trouble-host-unavailable.html.markerb
@@ -5,19 +5,21 @@ layout: docs
 nav: firecracker
 ---
 
-Fly.io has multiple hosts (physical hardware) in every supported region. A host could be unavailable because of a hardware failure, connectivity issues, or other reasons. A host issue might result in a complete or intermittent outage, or generally degraded performance for apps and databases without redundancy. A host could be down for an hour or a day, or sometimes longer. We understand that this can be frustrating if your app doesn’t have extra Machines or replicated databases for redundancy.
+Fly.io has multiple hosts (physical hardware) in every supported [region](/docs/reference/regions/). From time to time, a hardware failure, connectivity issues, or other factors will make a host unavailable. A single-host issue may cause a sustained or intermittent outage, or generally degraded performance, and this will disproportionately affect apps that run on a single Machine.
 
-For production apps, or any apps that require high availability, we strongly recommend running multiple machines and volumes for redundancy in the case of a single host failure. This goes double for apps with volumes, as volumes are pinned to specific hosts.
+A host can be down for an hour, or a day, or sometimes longer. 
+
+For production apps, or any apps that require high availability, we strongly recommend redundancy to protect against single-host failure. This means running multiple Machines, and for stateful apps it means having a clustering mechanism (such as those built into LiteFS and Postgres, for example) to replicate data between multiple volumes, because volumes are pinned to specific hosts.
 
 <div class="important icon">
-**Important:** The steps for recovery in this document don’t guarantee that you’ll be able to access your app if you have a single Machine&mdash;or a single Machine with an attached volume&mdash;on an a host that’s affected by an ongoing issue.
+**Important:** The steps for recovery in this document don’t guarantee that you’ll be able to access your app if you have a single Machine on a host that’s affected by an ongoing issue.
 </div>
 
 ## Notification of host issues
 
 You’ll get a notification on the Status page in your [dashboard](https://fly.io/dashboard/) when there’s a host issue that affects one or more of your apps. We don’t post individual host issues to our main status page because a host issue usually affects a small number of users in a single region.
 
-## For apps with one Machine and no volumes
+## Apps with one Machine and no volumes
 
 You can usually restore an app with one Machine and no volumes by scaling down and back up, or by deploying a new version of the app.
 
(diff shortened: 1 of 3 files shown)


apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
+---
+title: Troubleshoot apps when a host is unavailable
+objective: 
+layout: docs
+nav: firecracker
+---
+
+Fly.io has multiple hosts (physical hardware) in every supported region. A host could be unavailable because of a hardware failure, connectivity issues, or other reasons. A host issue might result in a complete or intermittent outage, or generally degraded performance for apps and databases without redundancy. A host could be down for an hour or a day, or sometimes longer. We understand that this can be frustrating if your app doesn’t have extra Machines or replicated databases for redundancy.
+
+For production apps, or any apps that require high availability, we strongly recommend running multiple machines and volumes for redundancy in the case of a single host failure. This goes double for apps with volumes, as volumes are pinned to specific hosts.
```
catflydotio, line 10:
```suggestion
Fly.io has multiple hosts (physical servers) in every supported [region](/docs/reference/regions/). From time to time, a hardware failure, connectivity issues, or other factor will make a host unavailable. A single-host issue may cause a sustained or intermittent outage, or generally degraded performance, and this will disproportionately affect apps that run on a single Machine.

A host can be down for an hour, or a day, or sometimes longer. 

For production apps, or any apps that require high availability, we strongly recommend redundancy as a buffer against single-host failure. This means running multiple Machines, and for stateful apps it means having a clustering mechanism (such as those built into LiteFS and Postgres, for example) to replicate data between multiple volumes. as volumes are pinned to specific hosts.
```

Take anything you like from here

apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
...
+
+<div class="important icon">
+**Important:** The steps for recovery in this document don’t guarantee that you’ll be able to access your app if you have a single Machine&mdash;or a single Machine with an attached volume&mdash;on an a host that’s affected by an ongoing issue.
```
catflydotio, line 13: I'm a bit confused by this sentence. Does this help at all?

```suggestion
**Important:** The steps for recovery in this document don’t guarantee that you’ll be able to access your app if you have a single Machine&mdash;or your data, if you have a single Machine with an attached volume&mdash;on an a host that’s affected by an ongoing issue.
```

apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
...
+
+2. Scale back up to 2 or more Machines:
+
+    ```cmd
+    fly scale count 2
+    ```
+
+<div class="note icon">
+**Note:** When you scale all your app's Machines to zero, the Machine size resets to shared-cpu-1x with 256MB of RAM. Use one of the `--vm-` options to size the Machines. For example: `fly scale count 2 --vm-size shared-cpu-4x`. See the [fly scale count docs](docs/flyctl/scale-count/) for options.
+</div>
```
catflydotio, line 42: Oh, I wonder how the `[compute]` section in `fly.toml` interacts with `fly scale count` (this is tangential, don't wait to merge)

apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
...
+
+
+### Deploy the app to get new Machines
+
+In most cases deploying a new version of your app with `fly deploy` should also work for an app without volumes, as the new machines will be deployed to healthy hosts.
```
catflydotio, line 47:
```suggestion
In most cases deploying a new version of your app with `fly deploy` should also work for an app without volumes, as the new Machines will be deployed to healthy hosts.
```

I'm not sure what's going on with Vale atm, but.

apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
...
+
+<div class="important icon">
+<b>Important:</b> Running a new deploy during a host outage generally won't destroy the existing Machine. When the impacted host returns, the Machine that was on it may return as well. In the case of a deploy during an outage, this can result in one Machine running an older version of your code. In that case you should destroy the outdated Machine if it comes back.
```
catflydotio, line 50: I don't know about anyone else but "deploy" sticks in my craw as a noun. That's a Heroku-ism, isn't it?

apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
...
+</div>
+
+## For apps with one Machine and an attached volume
```
catflydotio, line 53: Probably either all the headings should start with "for" or none should? I lean toward none if you need a tiebreaker but I can roll with all.

apps/trouble-host-unavailable.html.markerb
```diff
@@ -0,0 +1,83 @@
...
+
+If you're running a high availability Postgres cluster with multiple nodes, a host issue impacting one of your nodes shouldn't cause an issue; by default we run each node on a separate host. If the host your primary node is on goes down, the cluster will fail over to a healthy node. 
+
+However, if your database is running on a single Machine, and you don’t have any replicas to fail over to, then you won’t be able to connect during the host outage. Similar to the single volume steps above, you can create a new Postgres app from your most recent volume snapshot using `fly postgres create --snapshot-id <snapshot_id>`. See [Backup, restores, & snapshots](/docs/postgres/managing/backup-and-restore/) for details.
+
+## Prevent downtime when there's a single host issue
+
+Fly.io strongly recommends running at least two Machines per app in your primary region and we have features that can help make [app availability and resiliency](/docs/reference/app-availability/) more affordable.
+
+When possible, we place volumes on different hosts in the primary region. If one host goes down, we’ll fail over to the healthy host.
```
catflydotio, line 79: > If one host goes down, we’ll fail over to the healthy host.

I'm not sure what this refers to.

apps/trouble-host-unavailable.html.markerb
andie787: vale seems to die on old branches :(

apps/trouble-host-unavailable.html.markerb
andie787: > When possible, we place volumes on different hosts in the primary region. If one host goes down, then your app will still have a Machine and volume on a healthy host.

apps/trouble-host-unavailable.html.markerb
andie787: will look into this post-merge

PR #1094, opened 2023-10-08
Adding gem and rails installation guides

chethanuk:
n/a if none

### Notes
n/a if none


kcmartin:
Thanks for putting real effort into this, and sorry for the long delay in responding! This particular page is meant to be the fastest path to deploying on Fly.io, so it intentionally assumes Ruby/Rails are already installed and links out to the official Rails install guide rather than covering setup itself. A specific version-manager walkthrough (and the pinned Ruby 3.1.4, now end-of-life) is a bit out of scope here and would need ongoing maintenance. Going to close this one out, but genuinely appreciate the contribution.

PR #1105, opened 2023-10-11
Add Ruygt to linter accept list

catflydotio:
Add Ruygt to linter accept list, so Vale doesn't think it's a typo. (It's in some alt text for an image)


andie787:
LGTM!

PR #1112, opened 2023-10-15
Update install-flyctl.html.markerb

VinayKhatrii:
The changes i made seems to be working for latest version of powershell.

n/a if none

### Notes
n/a if none


andie787:
I think, for now, we are fine with having the note to try `powershell` if `pwsh` doesn't work.

PR #1114, opened 2023-10-16
Disperse "Working with fly apps" content

andie787:
### Summary of changes

Move or remove the content from "Working with Fly Apps" from the getting started section.

- "Check deployment status" is covered in [Hands-on](https://fly.io/docs/hands-on/), [Speedrun](https://fly.io/docs/speedrun/), and [Get information about an  app](https://fly.io/docs/apps/info/).
- "Set secrets" is covered in [Secrets and Fly Apps](https://fly.io/docs/reference/secrets/).  Updated "Secrets and Fly Apps" with a few tidbits from "Set secrets" and made some other minor edits.
- "Fly.io and custom domains", is probably covered by [Custom Domains and SSL Certificates](https://fly.io/docs/app-guides/custom-domains-with-fly/), however, that doc is really long and we don't have time to verify it, so I moved the custom domain content to a new doc under Fly Launch called "Use a custom domain". Made some minor edits to to the content. 
- Deleted "Working with Fly Apps"
- Updated nav

n/a

### Notes
TBD for secrets: we should probably have a separate how-to and reference for secrets at some point.

## Preview docs

https://aa-docs.fly.dev/docs/apps/custom-domain/
https://aa-docs.fly.dev/docs/reference/secrets/


reference/secrets.html.markerb
```diff
@@ -15,37 +15,36 @@ Secrets are stored in an encrypted vault. When you set a secret through flyctl,
...
 
-To set, or update, a secret in the app's vault, but defer updating the Machines to later, use the `--stage` option:
+To set, or update, a secret in the app's vault, but defer updating the Machines to later, use the `--stage` option. For example:
 
 ```cmd
 fly secrets set DATABASE_URL=postgres://example.com/mydb --stage
 ```
 
-In this case, the staged secret will be available only on Machines that are started or updated after the `fly secrets set` command was run.
+In the preceding example, the staged secret is available only on Machines that are started or updated after running the `fly secrets set` command.
```
catflydotio, line 36: This says the Machines ran the `fly secrets set` command, doesn't it?

andie787: oops yeah

catflydotio (approved):
Not that you need it but here's an approval

PR #1159, opened 2023-11-07
Add `concurrency` key to CD github action guide

dejagersh:
I often run into the situation where multiple PRs are approved & merged in quick succession. If PR 2 merges quickly after PR 1, then deploy will start for PR 1. Shortly after another deploy will start and it will likely fail because PR 1 is still running and it will crash because some Fly lock can't be obtained. Result: the Fly app will have the code from PR 1 and PR 2 will not be deployed until a future push.

This change makes sure that deploys for each successive commits are queued and will run immediately after each other, rather than concurrently.


andie787:
Was updating this doc with other things and also used your suggestion with a few minor changes. Thanks for pointing this out! #1271

PR #1170, opened 2023-11-13
A usage doc for fly machine run

catflydotio:
### Summary of changes

New doc about using `fly machine run`, and nav entries in firecracker and machines sidebar partials

Removes old "run machines with flyctl" from main nav. Should make all the info in there exist elsewhere in docs before actually deleting

### DIFF
--- a/machines/run.html.md
+++ b/machines/run.html.md
@@ -22,7 +22,7 @@ Here's what `fly machine run` does for you:
 
 To create a Machine, but not start it, use [`fly machine create`](/docs/flyctl/machine-create/).
 
-To make changes to a Machine once it's created or run, use [`fly machine update`](/docs/flyctl/machine-create/).
+To make changes to a Machine once it's created or run, use [`fly machine update`](/docs/flyctl/machine-update/).
 
 To create and run a new Machine with the same configuration as an existing Machine, use [`fly machine clone`](/docs/flyctl/machine-clone/).
 
--- a/machines/run.html.md
+++ b/machines/run.html.md
@@ -269,17 +269,21 @@ fly machine run . --name my-special-Machine
 
 ## Place data into files on the Machine
 
-Configuration can be used to place data into files on a Machine's file system. The `fly machine run` command has three ways to make use of this. There's a limit to how large a file you can create in this way.
+Configuration can be used to place a limited quantity of data into files on a Machine's file system. The `fly machine run` command has three ways to make use of this. 
+
+<div class="important icon">
+This won't work for large files, as there's a limit to how much data can be stored in an app secret or a Machine's configuration.
+</div>
 
 ### Copy a local file into the Machine file system
 
 If it's not convenient to build a file into the Machine's Docker image, use the `--file-local` flag to store the contents of a local file in its configuration instead. 
 
 ```
---file-local=/path/inside/machine=<local/path>
+--file-local /path/inside/machine=local/path
 ```
 
-The file's contents are stored as a Base64-encoded string as part of the Machine configuraton, and decoded to the original text on the Machine.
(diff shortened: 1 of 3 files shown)


machines/run.html.md
```diff
@@ -0,0 +1,368 @@
...
+```
+ 
+Use `--org <org-name>` to specify which organization a newly created app should belong to. The `--org` flag is ignored when creating the new Machine in an existing app.
+
+
+## Get or build the Docker image
+
+All Fly Machines are made from Docker images. When you `fly launch` an app, a Fly Launch scanner may generate one for you based on your source code, or it may be a prebuilt image you point to, or it may be built from your hand-crafted Dockerfile and source code.
+
+We're working more directly with Machines here, so your choices with `fly machine run` are "point to an image" or "point to a Dockerfile."
```
andie787, line 64: I found the introduction of "fly launch"ing an app a bit confusing. You can take or leave whatever from my suggestion below, which attempts to flip around the two scenarios so `fly m run` is first.
```suggestion
All Fly Machines are made from Docker images. When you work directly with Machines using `fly machine` commands, you can point to an prebuilt image, or build an image from your hand-crafted Dockerfile and source code.

When you `fly launch` an app, you can still point to image or point to Dockerfile, or a Fly Launch scanner may generate an image for you based on your source code.

```

machines/run.html.md
```diff
@@ -0,0 +1,368 @@
...
+```
+
+Use the `--dockerfile` option to specify a Dockerfile with a different name. 
+
+```cmd
+fly machine run . --dockerfile Dockerfile-dev
+```
+
+### Use an existing image
+
```
andie787, line 80: Add "For example:"? here and above?

machines/run.html.md
```diff
@@ -0,0 +1,368 @@
...
+exec "$@"
+```
+
+## Choose the region
+
+Tell the Fly Platform which [region](/docs/reference/regions/) to create the Machine in with the `--region` flag; if for some reason it can't start a new Machine in that region, you'll get an error. If the `--region` flag is omitted, the platform chooses the nearest region to you.
+
+
+## Set hardware specs
```
andie787, line 140: do we say hardware specs or machine resources here?

andie787 (commented):
This is a great summary of all the things!

Would it be useful to include a just a note about using `fly m clone` to create new Machines with the same config?

machines/run.html.md
catflydotio: I bet people will find it better with hardware specs, but machine resources is more accurate

machines/run.html.md
```diff
@@ -0,0 +1,368 @@
+---
+title: Run a new Machine
```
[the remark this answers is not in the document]
catflydotio: I take the point but I'm not sure. That seems like a lot of words. I'm leaving it for now and we can change it later.

machines/run.html.md
```diff
@@ -0,0 +1,375 @@
...
+* Checks with the platform for the org and app you've specified, if any, and if needed, guides you through naming a new app
+* Creates a Fly App, if applicable
+* Gets, or builds, a Docker image to make the Machine from
+* Creates the Machine with some default config, plus config you pass to it with flags
+* Starts the Machine
+* Waits for the Machine to start before declaring success (or failure)
+
+To create a Machine, but not start it, use [`fly machine create`](/docs/flyctl/machine-create/).
+
+To make changes to a Machine once it's created or run, use [`fly machine update`](/docs/flyctl/machine-create/).
```
catflydotio, line 25:
```suggestion
To make changes to a Machine once it's created or run, use [`fly machine update`](/docs/flyctl/machine-update/).
```

catflydotio:
> This is a great summary of all the things!
> 
> Would it be useful to include just a few words about using `fly m clone` to create new Machines with the same config?

Done!

PR #1178, opened 2023-11-15
Document `mounts.initial_size`

dangra:
* https://github.com/superfly/flyctl/pull/3023
* https://github.com/superfly/flyctl/pull/3031
* https://github.com/superfly/flyctl/pull/3038

### Notes
n/a if none

### DIFF
--- a/reference/configuration.html.markerb
+++ b/reference/configuration.html.markerb
@@ -592,7 +592,8 @@ This section supports the Volumes feature for persistent storage. The section ha
 [mounts]
   source = "myapp_data"
   destination = "/data"
-  processes= ["disk"] # optional - attach volumes to Machines that belong to one or more process groups
+  processes = ["disk"] # optional - attach volumes to Machines that belong to one or more process groups
+  initial_size = "20gb" # optional - the size of the volume created on first deploy
 ```
 
 ### `source`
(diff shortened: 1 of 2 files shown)


reference/configuration.html.markerb
```diff
@@ -607,6 +608,14 @@ The `destination` is the directory where the `source` volume should be mounted o
 
 Optionally, you can specify a list of [processes](#the-processes-section) to limit volume mounts by process group. You can even specify two different `[[mounts]]` sections to mount different source volumes to VMs in different process groups. Note the need to use double brackets if there are multiple `[[mounts]]` defined.
 
+### `initial_size`
+
+Optionally, you can specify the size of the volume to be created when launching an app for the first time.
+
+This value is only used by `fly launch` and `fly deploy` commands in case they need to create a new volume, it is ignored by `fly volume create` command.
```
andie787, line 615: nitpick:
```suggestion
This value is only used by `fly launch` and `fly deploy` commands in case they need to create a new volume; it is ignored by the `fly volume create` command.
```

reference/configuration.html.markerb
```diff
@@ -607,6 +608,14 @@ The `destination` is the directory where the `source` volume should be mounted o
...
+
+The value can be a string with units like in `initial_size = "30GB"` or an integer like in `initial_size = 30`, in which case GB is assumed as unit.
```
andie787, line 617: nitpick:
```suggestion
The value can be a string with units, like `initial_size = "30GB"` for example. The value can also be an integer, like `initial_size = 30` for example, with GB as the default unit.
```

andie787 (approved):
This is a great feature!
Left a couple of nitpick-y suggestions.

PR #1193, opened 2023-11-19
Fix litefs-cloud auth token command

2m:
### Summary of changes

Fixed the `fly` CLI command for getting LiteFS Cloud auth token.


2m:
Nah, that does not work either. :) I was not able to find a CLI command to generate a new token.

PR #1209, opened 2023-11-24
Edits to launch page for launch ui

andie787:
### Summary of changes

- update the Fly Launch page a bit to adjust for the launch ui
- some other wording changes for hopefully clarity, less words

n/a

### Notes
n/a

### DIFF
--- a/apps/launch.html.markerb
+++ b/apps/launch.html.markerb
@@ -7,25 +7,27 @@ order: 10
 redirect_from: /launch/
 ---
 
-With Fly Launch, you can create a brand new app on Fly.io by running this command from the source directory of your project:
+Fly Launch helps you configure and orchestrate your app. 
+
+To create a brand new app on Fly.io, run this command from the source directory of your project:
 
 ```cmd
 fly launch
 ``` 
 
-If you're in a hurry to try your first launch, we also have a condensed [Hands-on](/docs/hands-on/) that launches a super-simple demo app. You might also be interested in our [language- and framework-specific launch guides](/docs/languages-and-frameworks/).
+If you're in a hurry to try your first launch, we have a condensed [Hands-on](/docs/hands-on/) that launches a super-simple demo app. Or check out our [language- and framework-specific launch guides](/docs/languages-and-frameworks/).
 
 ## Ingredients for a successful `fly launch`
 
 Here are the components of a successful launch, ready for the first deployment:
 
-* a way to get a Docker image that Fly.io can use to boot up a Machine with your app running
-* Fly.io-specific [configuration](/docs/reference/configuration/) for the app, in the form of a `fly.toml` file
-* optional platform resources such as [public IP addresses](/docs/reference/services/), [Fly Volumes](/docs/reference/volumes), [app secrets](/docs/reference/secrets/), [Fly Postgres](/docs/postgres/) clusters, integrated resources like [Upstash for Redis](/docs/reference/redis/)
+* A Docker image that Fly Launch can use to boot up a Machine for your app to run on
+* A `fly.toml` file for your app's Fly.io-specific [configuration](/docs/reference/configuration/)
+* Optional platform resources such as [public IP addresses](/docs/reference/services/), [Fly Volumes](/docs/reference/volumes), [app secrets](/docs/reference/secrets/), [Fly Postgres](/docs/postgres/) clusters, and integrated resources like [Upstash for Redis](/docs/reference/redis/)
 
 ## Framework launch scanners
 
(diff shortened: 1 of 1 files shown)


apps/launch.html.markerb
```diff
@@ -7,73 +7,86 @@ order: 10
...
+If you're in a hurry to try your first launch, we have a condensed [Hands-on](/docs/hands-on/) that launches a super-simple demo app. Or check out our [language- and framework-specific launch guides](/docs/languages-and-frameworks/).
 
 ## Ingredients for a successful `fly launch`
 
 Here are the components of a successful launch, ready for the first deployment:
 
-* a way to get a Docker image that Fly.io can use to boot up a Machine with your app running
-* Fly.io-specific [configuration](/docs/reference/configuration/) for the app, in the form of a `fly.toml` file
-* optional platform resources such as [public IP addresses](/docs/reference/services/), [Fly Volumes](/docs/reference/volumes), [app secrets](/docs/reference/secrets/), [Fly Postgres](/docs/postgres/) clusters, integrated resources like [Upstash for Redis](/docs/reference/redis/)
+* A Docker image that Fly Launch can use to boot up a Machine for your app to run on.
```
catflydotio, line 24: The first deployment is often where the Docker image comes into being, which is the reason for my original awkward wording.

I think that instead of Fly Launch, we're looking at the Fly Platform? flyd? The Fly Launch feature helps you configure and build, but once you have an image, you can use an API call to boot up a Machine.

(I don't think we want to call `fly m run` part of the Fly Launch feature, but even that will take a Dockerfile and your source code and build you an image before starting a Machine)

apps/launch.html.markerb
```diff
@@ -7,73 +7,86 @@ order: 10
...
+* A `fly.toml` file for your app's Fly.io-specific [configuration](/docs/reference/configuration/).
```
catflydotio, line 25: I liked putting the "what it is" before "which file it's in," but if you like this better, that's 👍

apps/launch.html.markerb
```diff
@@ -7,73 +7,86 @@ order: 10
...
 
 ## Framework launch scanners
 
-Depending on your project, `fly launch` may be able to look at your app's source code and get through that ingredient list, straight to a ready-to-deploy Fly App. This is most likely for frameworks on which Fly.io has people specializing full time. Right now that's [Elixir/Phoenix](/docs/elixir/), [Laravel](/docs/laravel/), [Rails](/docs/rails/), and [Django](/docs/django/), but we also have launch guides for other [languages and frameworks](/docs/languages-and-frameworks/).
+Depending on your project, `fly launch` may be able to look at your app's source code and go straight to a ready-to-deploy Fly App. This is most likely to work for frameworks on which Fly.io has people specializing full time. Right now that's [Elixir/Phoenix](/docs/elixir/), [Laravel](/docs/laravel/), [Rails](/docs/rails/), and [Django](/docs/django/), but we also have launch guides for other [languages and frameworks](/docs/languages-and-frameworks/).
```
catflydotio, line 30: Is it worth finding a clear way to refer back to that list of "components of a successful launch", to make it clear what `fly launch` can do, and that it doesn't always manage that?

apps/launch.html.markerb
```diff
@@ -7,73 +7,86 @@ order: 10
...
 
 If `fly launch` doesn't have a scanner that can set up your app automatically, it will still initialize a new Fly App in your Fly.io organization and provide you with a default app configuration that's a reasonable starting point for a simple web app. 
 
-It's also possible to perform an entirely manual "launch", skipping all the launch scanners and full-service resource provisioning, using `fly apps create`, a hand-crafted (or copied) `fly.toml`, and step-by-step resource provisioning, followed by `fly deploy`.
+You can also perform an entirely manual "launch", skipping all the launch scanners and full-service resource provisioning, using `fly apps create`, a hand-crafted (or copied) `fly.toml`, and step-by-step resource provisioning, followed by `fly deploy`.
 
 ## After `fly launch`
 
-If you've run `fly launch` but haven't deployed yet, you can:
+If ran `fly launch` but didn't deploy yet (hint: you can do this with `fly launch --no-deploy`), or you deployed but want to make changes, then you can:
```
catflydotio, line 80: Is there a convention to replace the present perfect with the past simple in this kind of doc? I believe it's less correct in generic English.

andie787: I questioned this too, and obviously i forgot the "you". But there is no convention. Will look at it again.

apps/launch.html.markerb
andie787: but back some words and changed Fly Launch to "we" for now

PR #1214, opened 2023-11-29
Update volumes doc to show launchv2

IohannesArnold:
Not deploying as part of fly launch is now a commandline flag only, not a dialogue, so change /docs/apps/volume-storage/ to show that.

n/a if none

### Notes
n/a if none


apps/volume-storage.html.markerb
```diff
@@ -11,16 +11,15 @@ Fly Volumes are local persistent storage for [Fly Machines](/docs/machines/). Le
...
+    fly launch --no-deploy
     ```
 
     ```out
     ...
     Wrote config file fly.toml
-    ? Would you like to deploy now? No
     ...
     Your app is ready! Deploy with `flyctl deploy`
     ```
```
andie787, line 25: I think we can just remove the output code bock for now.

IohannesArnold: Done

PR #1217, opened 2023-11-30
Fix uncordon endpoint and description of a bluegreen deployment.

asib:
### Summary of changes

The endpoint for `uncordon` was wrong. Also noticed that the description for a bluegreen deployment isn't quite right, since we don't update existing machines in a bluegreen deployment (see linked doc).

- https://fly.io/docs/reference/configuration/#picking-a-deployment-strategy

### Notes
n/a


andie787 (approved):
Thanks you for finding and fixing! lgtm!

PR #1220, opened 2023-11-30
Add notes about on-fail vs on-failure to Machine restart policy doc

andie787:
### Summary of changes

- clarify when you need to use `on-failure` vs `on-fail`, API calls fail with `on-fail`

https://github.com/superfly/flyctl/issues/2844

### Notes
n/a


fideloper:
image

I'm enjoying this like I enjoy those word magnets you combine on your fridge

PR #1226, opened 2023-12-02
Typo in Continuous Deployment with GitHub

opiruyan:
In the speed-run section FLY_API_TOKEN is created in the step 5 but being referenced as created in step 4.

### Summary of changes
Change the number


andie787:
thanks for catching this!

PR #1245, opened 2023-12-19
Update docs on regional failover - set top-level primary_region as well

Roadmaster:
### Summary of changes
Mention that when doing a regional failover the top-level primary_region must be set - this was noted by a support customer and we verified this experimentally.

n/a

### Notes
On tests, if only the environment variable is set, then failover doesn't happen at all.

### DIFF
--- a/postgres/advanced-guides/high-availability-and-global-replication.html.markerb
+++ b/postgres/advanced-guides/high-availability-and-global-replication.html.markerb
@@ -53,7 +53,7 @@ If you haven't already pulled down your `fly.toml` configuration file, you can d
 fly config save --app <app-name>
 ```
 
-Now, let's open up the `fly.toml` file and set the `PRIMARY_REGION` [top-level config setting](https://fly.io/docs/reference/configuration/#primary-region) and environment variable to our target region. In this example, we are looking to move leadership into the `lax` region.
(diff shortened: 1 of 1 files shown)


postgres/advanced-guides/high-availability-and-global-replication.html.markerb
```diff
@@ -46,15 +46,17 @@ Failover complete
...
-There may be situations where you want to move leadership into a completely new region.  While this process is a bit more involved, you can achieve this by performing the steps below.  
+There may be situations where you want to move leadership into a completely new region.  While this process is a bit more involved, you can achieve this by performing the steps below.
 
 If you haven't already pulled down your `fly.toml` configuration file, you can do so by running:
 ```
 fly config save --app <app-name>
 ```
 
-Now, let's open up the `fly.toml` file and set the `PRIMARY_REGION` environment variable to our target region. In this example, we are looking to move leadership into the `lax` region. 
+Now, let's open up the `fly.toml` file and set the `PRIMARY_REGION` [top-level config setting](https://fly.io/docs/reference/configuration/#primary-region) and environment variable to our target region. In this example, we are looking to move leadership into the `lax` region.
```
andie787, line 56:
```suggestion
Now, let's open up the `fly.toml` file and set the [`primary_region`](https://fly.io/docs/reference/configuration/#primary-region) and the `PRIMARY_REGION` environment variable to our target region. In this example, we are looking to move leadership into the `lax` region.
```

andie787 (approved):
Thanks for this update. There is a further update in the works and this info is helpful! Left one suggestion.

PR #1249, opened 2023-12-21
Update _launch_with_postgres.html.erb

rahul190201:
Please refer to the following issue:
https://community.fly.io/t/problem-launch/17319?u=rahul190201

n/a if none

### Notes
n/a if none


andie787:
Thanks for submitting a workaround. We've made some changes to fix this issue, so that 'fly launch' doesn't attempt to launch apps in the maa region (which has been closed).

PR #1251, opened 2023-12-22
Fix broken link to private networking docs

ab:
I think these docs have moved to their own page.

### Summary of changes
Fix link to 6PN naming docs page.


andie787 (approved):
Thanks for the fix!

PR #1254, opened 2023-12-27
Clarify app name needs to be unique

kylemclaren:
### Summary of changes

Add statement to clarify that app names need to be unique across all of Fly. Not immediately obvious to newcomers and launch can fail in weird/unexpected ways.


andie787 (approved):
lgtm!

PR #1255, opened 2023-12-27
Add a callout to Supabase docs for private alpha

jsierles:


kcmartin:
Closing this one out: Supabase Managed Postgres on Fly.io has since been discontinued, and the docs now point to the shutdown announcement, so the alpha signup callout is no longer relevant.

PR #1258, opened 2024-01-03
Clarify that min_machines_running doesn't apply to non-primary regions

jsierles:
The previous copy was confusing:

> The `min_machines_running` value applies to the total number of Machines, not Machines per region. 

This could be interpreted as meaning 'total number of machines across all regions'.


andie787:
lgtm! thanks!

jsierles:
I'm not sure this is accurate, so going to leave it until confirming the behavior.

andie787:
sounds good. It _is_ the expected behaviour, but doesn't hurt to confirm.

PR #1260, opened 2024-01-04
update livebook image version

afomi:
update livebook image version


andie787:
Could we use `livebook:latest`?

kcmartin:
Thanks for the bump, and sorry for the long delay! The guide has since moved past this (currently pins 0.13.0), so this specific change is superseded.

PR #1263, opened 2024-01-05
Update consul install instructions for current debian releases

fdietze:
The `consul` package is not available anymore in Debian bookworm.


kcmartin (approved):
Sorry for the long delay. This is still needed. Approving and merging in.

PR #1268, opened 2024-01-07
Minor spelling mistake

YandeMC:
fix a single spelling mistake

### Summary of changes
replaced "an" with "and" in "any number between 1024 an 65535."

thats it.


kylemclaren (commented):
Nice catch. I think "application" should be pluralized too.

PR #1272, opened 2024-01-08
Shared IPv4 needs a cert for custom domain

catflydotio:
### Summary of changes
When an app uses shared IPv4, Fly Proxy needs an SSL cert to know which app to route a custom domain address to, even if HTTPS isn't used. This PR adds this info to:
* https://fly.io/docs/reference/services/ (instructions for moving to shared ipv4)
* https://fly.io/docs/apps/custom-domain/
* https://fly.io/docs/app-guides/custom-domains-with-fly/

Also did minor cleanup and corrected out-of-date statement that allocating shared IPv4 would fail if dedicated IPv4 was present.

https://community.fly.io/t/new-shared-ipv4-connection-reset-by-peer/17567

### DIFF
--- a/reference/services.html.markerb
+++ b/reference/services.html.markerb
@@ -55,14 +55,17 @@ IPv6 addresses and shared IPv4 anycast addresses are free. Dedicated IPv4 addres
 
 ### Switch from dedicated IPv4 to shared IPv4
 
-The following steps should allow you to switch an app with a custom domain from a dedicated IPv4 address to a free shared IPv4 without downtime:
+The following steps should allow you to switch an app with a custom domain from a dedicated IPv4 address to a free shared IPv4 without downtime. 
 
-1. Add shared IPv4 via `flyctl ips allocate-v4 --shared`
-2. If you don't use a CNAME to `.fly.dev`, [add the shared ipv4 as an A record](https://fly.io/docs/apps/custom-domain/#set-the-a-record).
+1. Add a shared IPv4 address via `flyctl ips allocate-v4 --shared`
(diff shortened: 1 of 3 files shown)


reference/services.html.markerb
```diff
@@ -55,14 +53,17 @@ IPv6 addresses and shared IPv4 anycast addresses are free. Dedicated IPv4 addres
...
-5. Remove the dedicated IPv4 from the app using `fly ips release <ipv4-address-to-release>`. You can view your app's IP addresses using `fly ips list`.
-6. Remove the unwanted IPv4 address from your DNS if it was set manually as an A record.
+4. If you are using a custom domain and don't already have an SSL certificate for this app, [create one](https://fly.io/docs/apps/custom-domain/#get-certified). The cert is required for routing to your app via the custom domain, even for HTTP connections.
+5. Wait for DNS caches to clear. Five minutes is likely enough, but this varies wildly. This is determined by the larger of your DNS record’s TTL and that of our `<your-app>.fly.dev` record.
+6. Remove the dedicated IPv4 from the app using `fly ips release <ipv4-address-to-release>`. You can view your app's IP addresses using `fly ips list`.
+7. Remove the unwanted IPv4 address from your DNS if it was set manually as an A record.
+
+If you do not have a custom domain (your app is accessed by its `.fly.dev` address), you don't have to change any DNS records yourself, and you don't need to add your own SSL cert.
```
andie787, line 66: Should this be a note at the top of the sub-section? And would they just do steps 1, 3, 6, and 7? :)

andie787 (commented):
One comment only. This is nice update!

pborzenkov (approved):
Looks good. Thank you!

reference/services.html.markerb
catflydotio: I'm merging as-is, to get the new info in; we can pick this up for improvements as priority allows

PR #1282, opened 2024-01-12
Update UDP service doc to include the IPv6 rule does not apply to 6PN

gargakshit:
### Summary of changes

One of our customers at the community forum wanted to use UDP over 6PN for app communication, and was confused about it working over 6PN or not as UDP services do not IPv6.

https://community.fly.io/t/internal-app-to-app-or-machine-to-machine-udp-communication/17533

### Notes
n/a

### DIFF
--- a/app-guides/udp-and-tcp.html.markerb
+++ b/app-guides/udp-and-tcp.html.markerb
@@ -176,6 +176,10 @@ Another thing that won't impact your code that you'll need to know about is that
 
 Every Fly.io app gets an IPv6 address, and that address will work fine for the TCP side of your app, for whatever that's worth.
 
+<div class="note icon">
+This does not apply to 6PN; UDP over 6PN works just as you would expect.
+</div>
+
 ### You Might Need To Be Mindful Of MTUs
 
 An invariant on almost every network is the maximum packet size, the MTU. Typically, the MTU is 1500 bytes. If you exceed 1500 bytes, your packets will fragment. [You don't want to fragment](https://datatracker.ietf.org/doc/html/draft-mathis-frag-harmful-00).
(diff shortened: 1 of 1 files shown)


app-guides/udp-and-tcp.html.markerb
```diff
@@ -176,6 +176,10 @@ Another thing that won't impact your code that you'll need to know about is that
 
 Every Fly.io app gets an IPv6 address, and that address will work fine for the TCP side of your app, for whatever that's worth.
 
+<div class="note icon">
+This does not apply to 6PN; UDP over 6PN works just as you would expect.
```
andie787, line 180: Suggest adding the word "note", and maybe including a link to 6PN info and removing the "just as you would expect" since they might not know about 6PN yet? Something like this (or whatever wording you think works!):
```suggestion
**Note:** UDP does work over private [6PN networks](/docs/reference/private-networking/).
```

andie787 (commented):
thanks for this update! Left a couple of suggestions for you

app-guides/udp-and-tcp.html.markerb
gargakshit:
```suggestion
**Note:** UDP does work over private [Fly IPv6 private networking](/docs/reference/private-networking/).
```

andie787: sorry being a pain, but if you remove the first "private" and change "Fly" to "Fly.io", then you can go ahead and merge away!

gargakshit:
```suggestion
**Note:** UDP does work over [Fly.io IPv6 private networking](/docs/reference/private-networking/).
```

Thanks!

PR #1304, opened 2024-01-19
Update supercronic app guide

schrockwell:
### Summary of changes
- Updated supercronic link for the latest version
- Fixed `fly scale count` command syntax

n/a

### Notes
The `fly scale count` syntax is taken from the current version of `flyctl`, `v0.1.143`

### DIFF
--- a/app-guides/supercronic.html.md
+++ b/app-guides/supercronic.html.md
@@ -32,13 +32,13 @@ Check out [cron.help](https://cron.help) if you need a quick crontab syntax refe
 
 ## Install `supercronic` in the container
 
-The [latest releases for supercronic](https://github.com/aptible/supercronic/releases) are on [Github](https://github.com/aptible/supercronic), where they include copy pasta 🍝instructions for getting it in your Dockerfile. As of November 2022, the current version of `supercronic` is v0.2.1. You'll want to check the [releases page](https://github.com/aptible/supercronic/releases) for the latest version, but here's what it looks like now:
+The [latest releases for supercronic](https://github.com/aptible/supercronic/releases) are on [Github](https://github.com/aptible/supercronic), where they include copy pasta 🍝instructions for getting it in your Dockerfile. As of January 2024, the current version of `supercronic` is v0.2.29. You'll want to check the [releases page](https://github.com/aptible/supercronic/releases) for the latest version, but here's what it looks like now:
 
 ```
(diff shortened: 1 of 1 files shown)


app-guides/supercronic.html.md
```diff
@@ -82,10 +82,10 @@ Now that we've added `supercronic` to our `Dockerfile`, put the `crontab` at the
 $ fly deploy
 ```
 
-Then we'll need to scale the processes so that we only run one virtual machine container with the `cron` process. Be sure to change `web` to whatever number you had before.
+Then we'll need to scale the processes so that we only run one virtual machine container with the `cron` process.
```
andie787, line 85: I'm not sure we should delete the part about the `web` process since this guide includes creating it that process (I assume it replaces the default `app` process present in a basic app.

app-guides/supercronic.html.md
```diff
@@ -82,10 +82,10 @@ Now that we've added `supercronic` to our `Dockerfile`, put the `crontab` at the
...
 
 ```
-$ fly scale count cron=1 web=
+$ fly scale count -g cron 1
```
andie787, line 88: This will work to scale one process group, but to scale more than one process group in a single command you would use the `fly scale count cron=1 web=2` for example.

andie787 (commented):
Thanks for the update. The version update looks good, but I'm not sure about the command updates.

app-guides/supercronic.html.md
schrockwell: This was the error I got when running the example command.

```
❯ fly scale count cron=1 web=2
Error: --process-group flag is required when no group named 'app' is defined
```

So I thought maybe the command was out-of-date. Maybe the entire page should change `web` to `app` since that seems to be the new convention?

andie787: Oh wow, yeah, that's not right. I'll look into things and then we'll see whether it's docs or the command that gets fixed. :)

andie787:
flyctl issue: https://github.com/superfly/flyctl/issues/2557

andie787:
@schrockwell since I fix in flyctl isn't going to happen right away, my solution would be to use `app` instead of `web` or to show running the `scale count` command for each group (or adding words to say repeat for each group).

andie787:
Fixed by https://github.com/superfly/flyctl/pull/3280

andie787:
made the version update you suggested in #1356

PR #1305, opened 2024-01-19
Mach pricing words

catflydotio:
n/a if none

### Notes
n/a if none

### DIFF
--- a/about/pricing.html.markerb
+++ b/about/pricing.html.markerb
@@ -7,13 +7,13 @@ nav: firecracker
 
 <img src="/static/images/app-pricing.webp" srcset="/static/images/app-pricing@2x.webp 2x" alt="A landscape painting with hot air balloons flying over a mountain range.">
 
-## _How it works_
+## How it works
 
 Our pricing is designed to let you launch a small application for free, and scale costs affordably as your needs grow.
 
 Fly.io services are billed per organization. Organizations are administrative entities on Fly.io that enable you to add members, share app development environments, and manage billing separately. Billing is based on the resources provisioned for your apps, pro-rated for the time they are provisioned. Learn more about [billing](/docs/about/billing/).
 
-## _Plans_
+## Plans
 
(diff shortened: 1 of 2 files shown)


about/pricing.html.markerb
```diff
@@ -39,11 +39,13 @@ Resources included for free on all plans:
...
+## Compute
 
 ### Fly Machines
 
-Pricing for [Fly Machines](/docs/machines/) VMs, which are also the virtual-machine building blocks of [Fly Launch](/docs/apps/) and [Fly Postgres](/docs/postgres/) apps.
+The price of a [Fly Machine](/docs/machines/) VM is the price of a preset CPU/RAM combination, plus about $5 per 30 days per GB of additional RAM.
+
+The available CPU/RAM combinations and their pricing are as follows:
 
 <%= partial("shared/cpu_mem_machines_pricing") %>
```
asib, line 50: Will this still show the current matrix? Because it includes combos that aren't "presets" as defined above. The only presets are:

shared-1x-256mb, shared-2x-512mb, shared-4x-1gb, shared-8x-2gb
perf-1x-2gb, perf-2x-4gb, perf-4x-8gb, perf-8x-16gb, perf-16x-32gb

jphenow: I can see reading [the matrix](https://fly.io//docs/about/pricing/#compute) as either "each row is a preset" or "the first of each grouping is the preset."

I think you reach the same bottom-line either way, but maybe if the preset => base RAM is very obvious (or the only thing on the matrix) the invoices will connect a little more.

I'm not sure!

jphenow (approved):

about/pricing.html.markerb
catflydotio: How about now? I put "named" in front of "presets" and made it explicit that the additional RAM configs are in the table.

catflydotio:
Merged. If there's an obvious improvement we can make from here, PR welcome :D

PR #1306, opened 2024-01-20
regions.html.markerb: fix typo

Goorzhel:
"Paolo" is Italian.


catflydotio:
Nice, thanks @Goorzhel

PR #1317, opened 2024-02-01
Updates to mySQL doc

andie787:
### Summary of changes

- moved specifying the image to the `fly launch` command
- removed the bit about the bug in 8.0.33
- added steps re: the launch UI
- moved the bit about scaling memory to the launch UI step (a user noticed that you could not use `fly scale` before the first deploy)
- other minor edits

- https://community.fly.io/t/scaling-memory-does-not-seem-to-work/6764/14

### Notes
n/a

### DIFF
--- a/app-guides/mysql-on-fly.html.markerb
+++ b/app-guides/mysql-on-fly.html.markerb
@@ -27,12 +27,17 @@ mkdir my-mysql
 cd my-mysql
 
 # Run `fly launch` to create an app
-fly launch --no-deploy
+# Use the --no-deploy option since we'll make changes before first deploy
+# Use the --image option to specify a MySQL docker image
+ly launch --no-deploy --image mysql:8
 ```
 
-You can name the app whatever you'd like. The name will become a hostname our application uses to connect to the database, such as `my-mysql.internal`.
+Type `y` when prompted to tweak the default app settings. Then, on the Fly Launch page:
 
-We used the `--no-deploy` option because we have some work to do before we deploy the app.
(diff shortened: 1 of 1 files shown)


app-guides/mysql-on-fly.html.markerb
```diff
@@ -59,39 +64,38 @@ Finally, edit the `fly.toml` file that `fly launch` generated.
...
+
+# Add the following sections
 [processes]
-app = """--datadir /data/mysql \
-  --default-authentication-plugin mysql_native_password \
-  --performance-schema=OFF \
-  --innodb-buffer-pool-size 64M"""
+  app = """--datadir /data/mysql \
+    --default-authentication-plugin mysql_native_password \
+    --performance-schema=OFF \
```
fideloper, line 84: The two options:

```
    --performance-schema=OFF \
    --innodb-buffer-pool-size 64M
```

...were to get MySQL 8 to work in low-RAM environments (e.g. `256m`), but the docs now assume users will run ~2gb of ram, making these flags unnecessary (probably a great idea).

The easiest thing to do is just remove the flags:

```
  app = """--datadir /data/mysql \
    --default-authentication-plugin mysql_native_password \
    --performance-schema=OFF \
```

The slightly harder version is  making a note to say if you use less than 1gb of RAM, perhaps use those last 2 flags. I'm not entirely sure where that fits best (in comments of fly.toml? as a new paragraph?

andie787: @fideloper oh was that to get it to deploy before scaling?

I'm confused about `--performance-schema`, should that be removed as well?

PR #1346, opened 2024-02-13
Document "gpus = N" for multi GPU support

dangra:
### Summary of changes

Document `gpus` fly.toml field  

* https://github.com/superfly/flyctl/pull/3199

### Notes
n/a if none

### DIFF
--- a/reference/configuration.html.markerb
+++ b/reference/configuration.html.markerb
@@ -683,9 +683,13 @@ The kind of CPU to request. Valid values are `shared` and `performance`.
 
 The number of GPUs to request. Valid values are `1`, `2`, `4`, `8`, but depends on `gpu_kind`. 
 
+Setting this value requires to also set `gpu_kind`.
+
 ### `gpu_kind`
 
 The kind of GPU to request. Valid values are `a100-pcie-40gb`, `a100-sxm4-80gb` and `l40s`.
+
+Setting this value implies `gpus = 1` if not present.
 
 ### `host_dedication_id`
 
--- a/reference/configuration.html.markerb
+++ b/reference/configuration.html.markerb
@@ -655,6 +655,7 @@ All keys are optional and `size` has lower precedence than all other keys.
   memory = "2gb"
   cpus = 4
   cpu_kind = "performance"
+  gpus = 1
   gpu_kind = "a100-pcie-40gb"
   host_dedication_id = "customer-id"
   processes = ["app"]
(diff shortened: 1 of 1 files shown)


reference/configuration.html.markerb
```diff
@@ -672,16 +673,24 @@ The memory to request. It can be expressed with units, like `memory = "2GB"` or
...
 
 ### `cpu_kind`
 
 The kind of CPU to request. Valid values are `shared` and `performance`.
 
+### `gpus`
+
+The number of GPUs to request. Valid values are `1`, `2`, `4`, `8`, but depends on `gpu_kind`. 
+
+Setting this value requires to also set `gpu_kind`.
```
andie787, line 686:
```suggestion
Setting this value requires also setting `gpu_kind`.
```

reference/configuration.html.markerb
```diff
@@ -672,16 +673,24 @@ The memory to request. It can be expressed with units, like `memory = "2GB"` or
...
-The number of vCPUs to request. Valid values are `1`, `2`, `4`, `8`, or `16`, but depends on `gpu_kind`. 
+The number of vCPUs to request. Valid values are `1`, `2`, `4`, `8`, or `16`, but depends on `cpu_kind`. 
 
 ### `cpu_kind`
 
 The kind of CPU to request. Valid values are `shared` and `performance`.
 
+### `gpus`
+
+The number of GPUs to request. Valid values are `1`, `2`, `4`, `8`, but depends on `gpu_kind`.
```
andie787, line 684: should we say to run `fly platform vm-sizes` to check? Or are those valid in this context? Current output for GPUs is:
```
NAME     	CPU CORES	MEMORY	GPU MODEL
a100-40gb	8        	32 GB 	a100-pcie-40gb
a100-80gb	8        	32 GB 	a100-sxm4-80gb
l40s     	8        	32 GB 	l40s
```
Should CPU be CUDA?

reference/configuration.html.markerb
```diff
@@ -672,16 +673,24 @@ The memory to request. It can be expressed with units, like `memory = "2GB"` or
...
+
+The number of GPUs to request. Valid values are `1`, `2`, `4`, `8`, but depends on `gpu_kind`. 
+
+Setting this value requires to also set `gpu_kind`.
+
 ### `gpu_kind`
 
 The kind of GPU to request. Valid values are `a100-pcie-40gb`, `a100-sxm4-80gb` and `l40s`.
 
+Setting this value implies `gpus = 1` if not present.
```
andie787, line 692:
```suggestion
Setting this value assumes `gpus = 1` if not present.
```

andie787 (commented):
left some questions and a few suggestions!

reference/configuration.html.markerb
dangra: GPU machines can get different _performance_ CPUs too (16x is possible for example, "shared" is not possible).
The confusion I think comes from `gpus` flag is not about how many CUDA cores but how many whole nvidia devices are allocated to the machine.

For instance the `l40s` machine size expands to `performance-8x` with 32gb ram and _1 x l40s_ GPU device.

What is missing in `fly platform vm-sizes` or any place where we show size aliases for GPU machines is a way to show how many GPU devices are allocated.

hopes that explains it 😕

PR #1352, opened 2024-02-15
Update install-flyctl.html.markerb

anadon:
Added `bash` and bash-like shell specific modifications required to directly invoke Fly.io commands.  Input taken from https://stackoverflow.com/a/30387689/826573 for prepending to text files.

For very new people, this kind of thing can trip them up.

n/a if none

### Notes
One odd part is prepending to `.bashrc`.  This is to not conflict with `direnv`.

### DIFF
--- a/hands-on/install-flyctl.html.markerb
+++ b/hands-on/install-flyctl.html.markerb
@@ -30,6 +30,7 @@ If not, you can run the install script:
 
 ```cmd
 curl -L https://fly.io/install.sh | sh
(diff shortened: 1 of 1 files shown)


kcmartin:
Thanks for the suggestion, and sorry for the long delay! This page has since been consolidated into `/docs/flyctl/install/`, which points users to the install script's own output for shell-appropriate rc entries. Closing, but appreciate the thought for new users!

PR #1394, opened 2024-03-02
Update autostart-stop.html.markerb

xHomu:
### Summary of changes

Match default mentioned in docs/reference/configuration


andie787:
hi @xHomu! Thanks for noticing the discrepancy in docs. The fix is actually the opposite; the fly.toml config reference was incorrect. Fixed in #1404

andie787:
#1404

PR #1401, opened 2024-03-06
update what we support/do not support

senyosimpson:
### Summary of changes

Updates to show what we do and don't support and what we have no plan in supporting

f.y.i @jipperinbham 

### Notes

andie787: Also temporarily removing the volumes doc while we work on updating it. Added a redirect to index and renamed the file to avoid visibility temporarily.


kubernetes/index.html.markerb
```diff
@@ -47,13 +47,20 @@ Connect to your cluster using kubectl. Clients are authenticated with our API to
 
 ### Not supported
 
-  * Persistent volumes (coming soon)
-  * Sidecars and init processes (coming soon)
+There are a few features we don't support yet but are part of the roadmap
```
andie787, line 50:
```suggestion
There are a few features we don't support yet but are part of the roadmap:
```

kubernetes/index.html.markerb
```diff
@@ -47,13 +47,20 @@ Connect to your cluster using kubectl. Clients are authenticated with our API to
...
+
+  * Sidecars and init processes
+  * EmptyDir volumes
   * Horizontal pod autoscaling
-  * Gateway API
+  * StatefulSets
+  * CronJob
+
+Features we don't support
```
andie787, line 58:
```suggestion
Features we don't support:
```

andie787 (commented):
The tiniest of changes and you're good to go!

PR #1403, opened 2024-03-06
Update the Machines API page: Remove "more performant" descriptor for public api, Mention less latency for Private api when accessed from Machine 

KTanAug21:
### Summary of changes

1. Removed "more performant" description of public Machine api as an alternative to the private Machine api,  This description might give the impression that the public api performs better than the private api.

2. Additionally mentioned that accessing the private Machine Api has less latency when accessed from a Fly Machine.

### Notes
There is less latency when connecting to the private api from a Fly Machine:
https://flyio.discourse.team/t/best-practices-for-public-vs-private-machines-api-endpoint/5019

### DIFF
--- a/machines/api/working-with-machines-api.html.markerb
+++ b/machines/api/working-with-machines-api.html.markerb
@@ -18,13 +18,13 @@ This guide assumes that you have `flyctl` and `curl` installed, and have authent
 
 ### Using the public `api.machines.dev` endpoint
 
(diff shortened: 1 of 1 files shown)


andie787 (commented):
This lgtm!

KTanAug21:
> This lgtm!

Thanks for checking @andie787 !!

PR #1410, opened 2024-03-08
Add tigirs to database storage guides

guissalustiano:
### Summary of changes
- Add tigirs to database storage guides


andie787:
thanks! needed to add another partner as well so this is fixed in #1411

PR #1413, opened 2024-03-08
fix broken postgres configuration link

montasaurus:
### Summary of changes

The link currently points to a section that doesn't exist.


andie787:
thanks for noticing and fixing!

PR #1418, opened 2024-03-10
Update private-networking.html.md as per https://bugs.launchpad.net/u…

anadon:
…buntu/+source/resolvconf/+bug/1990743

Updating install instructions to be more precise, incorporating information found in https://bugs.launchpad.net/ubuntu/+source/resolvconf/+bug/1990743 .


andie787:
I agree this is more specific, but I think Ubuntu 18 is specific enough, and providing 22.04 as the end of the range is misleading, because although 22.04 is the latest LTS, the statement implies that `openresolv` isn't required in 23.x releases. Is that the case?

anadon:
Not only it is not required, it is not available to install.

PR #1424, opened 2024-03-11
document configuring service concurrency

senyosimpson:
### Summary of changes

Add documentation on configuring Fly.io concurrency limits when using K8s services

### DIFF
--- a/kubernetes/services.html.markerb
+++ b/kubernetes/services.html.markerb
@@ -69,6 +69,31 @@ spec:
       targetPort: 8080
 ```
 
+## Concurrency limits
+
+[Concurrency limits](https://fly.io/docs/reference/configuration/#services-concurrency) are used to limit the load on your application.
+By default, the soft limit is set to 20 and the hard limit is set to 25.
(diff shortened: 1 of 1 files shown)


senyosimpson:
@jipperinbham mind just sanity checking the example

jipperinbham (approved):

kubernetes/services.html.markerb
```diff
@@ -69,6 +69,31 @@ spec:
       targetPort: 8080
 ```
 
+## Concurrency limits
+
+[Concurrency limits](https://fly.io/docs/reference/configuration/#services-concurrency) are used to limit the load on your application.
```
andie787, line 74: In this case, with a cluster having one Machines, there's no load balancing and no auto stop or start, right?

andie787 (commented):
lgtm! left one question

kubernetes/services.html.markerb
senyosimpson: yep! but if your one instance hits its hard limit, we'll just fail requests to your app entirely

Issue #1383, opened 2024-02-27
Private Networking - implies tunnel is up without running wg-quick


colemickens:
Look at the "Dealing with Defaults" section. It shows the tunnel working before the docs have the user run `wg-quick`.

Looks like maybe a mistake from moving the wg-quick instructions to being OS-specific and not (re)moving the the part where you query DNS to show it working after `wg-quick up`.

![Uploading screenshot-1709062220.png…]()

colemickens:
Brilliant work there GitHub. 
![screenshot-1709062220](https://github.com/superfly/docs/assets/327028/4debf7fd-c957-43af-820b-448448023938)

andie787:
Oh good point! Looks like this doc has been structured this way forever. It could probably use a review/update, so we'll take a look.

andie787:
I updated most of the doc, numbered the main steps in Private Network VPN and turned that section into a note and got rid of the "dealing with defaults" title altogether. Moved the dig examples into the Test the VPN tunnel section.

PR #1430, opened 2024-03-14
Private networking doc: update for Machines, clarify .internal addresses, dealing with defaults section

andie787:
### Summary of changes

- petty much rewrote the Fly.io .internal addresses section, to clarify how to use them and what they return
- added a note about .internal hostnames not returning info on stopped Machines
- rewrote all the TXT and AAAA descriptions in the table for clarity about what is actually returned on query
- re-organized the Provate network VPN section, after a user noted that the "Dealing with defaults" section was likely in the wrong place in the steps or made an incorrect assumption about that.
- format/update notes, headings, asides
- change "instances" to apps or Machines

https://github.com/superfly/docs/issues/1383

### DIFF
--- a/networking/private-networking.html.md
+++ b/networking/private-networking.html.md
@@ -16,9 +16,9 @@ This connectivity is always available to apps; you don't have to do anything spe
 
 You can connect apps running outside of Fly.io to your 6PN network using WireGuard; for that matter, you can connect your dev laptop to your 6PN network. To do that, you'll use flyctl, the Fly.io CLI, to generate a WireGuard configuration that is addressed with a 6PN address.
 
-## Discovering Apps through DNS on a Fly Machine
+## Discover apps through DNS on a Fly Machine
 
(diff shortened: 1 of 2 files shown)


networking/private-networking.html.md
```diff
@@ -134,7 +144,7 @@ VERSION	IP                	TYPE   	REGION	CREATED AT
 v6     	fdaa:0:22b7:0:1::3	private	global	just now
 ```
 
-If you want to expose services to another Fly organization you have access to, use the `--org` flag.
+If you want to expose services to another Fl.io organization you have access to, then use the `--org` flag.
```
catflydotio, line 147:
```suggestion
If you want to expose services to another Fly.io organization you have access to, then use the `--org` flag.
```

catflydotio (changes requested):

catflydotio (approved):

PR #1434, opened 2024-03-20
docs: update multiple-processes docs with correct internal DNS addresses

Drakeoon:
### Summary of changes

This PR adds small change to the "Multiple processes" section in the docs. Since internal domains like `foo.internal` and `bar.internal` are missing the`<region>.` part, I updated the description with the correct addresses.

This would help users come up with the proper internal DNS addresses without having to read other doc pages

- Changes directly update this page: https://fly.io/docs/app-guides/multiple-processes/
- Documentation with correct addresses: https://fly.io/docs/networking/private-networking/


Drakeoon:
Closed since the guide mentions `<app-name>.internal` addresses should work and pick up any deployed region

PR #1444, opened 2024-03-29
Clean up working with Machines API page

catflydotio:
### Summary of changes
* don't link to the external API app at the top
* don't recommend `fly machines api-proxy` since we have a public API URL now
* you don't have to use org-wide tokens -- didn't explain all the options; for now making it clear we're using the token flyctl is using in the example
* general tidying

### Notes
* still should explain somewhere about deploy tokens (better, have a doc on tokens)

### DIFF
--- a/machines/api/working-with-machines-api.html.markerb
+++ b/machines/api/working-with-machines-api.html.markerb
@@ -6,82 +6,51 @@ nav: machines
 order: 1
 ---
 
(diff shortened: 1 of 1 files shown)


catflydotio:
Start of rough phrasing for tokens stuff in abandoned branch https://github.com/superfly/docs/tree/c-using-api

PR #1450, opened 2024-04-03
Hostname validation will fail without an IPv6 address

kylemclaren:
### Summary of changes

In my testing I came across this issue — we perform a hostname check and unless an IPv6 is associated with an app, we won't attempt to issue (or renew) a cert.

(These docs are in need of a bigger overhaul, I think)

### Preview

Screenshot 2024-04-03 at 16 07 48

### DIFF
--- a/networking/custom-domains-with-fly.html.md
+++ b/networking/custom-domains-with-fly.html.md
@@ -31,13 +31,13 @@ There's a question to ask and answer. Do you want to start accepting traffic imm
 
 ### Accepting traffic immediately for the custom domain
 
-In this scenario, we want the custom domain to point to the `nginxproxy` server which will allow unencrypted IPv4 and IPv6 connections. Again, there are two ways to do this. Using DNS's CNAME capability or setting the A and AAAA records.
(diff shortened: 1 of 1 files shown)


networking/custom-domains-with-fly.html.md
```diff
@@ -54,6 +54,10 @@ fly ips list
 
 Create an A record pointing to your v4 address, and an AAAA record pointing to your v6 address. You're then free to make this an Apex domain as needed.
 
+<div class="important">
+Our hostname validation will fail without a v6 address and we won't attempt to issue or renew a certificate. If your app does not have one, you can allocate one with `flyctl ips allocate-v6`. If you use a CNAME `_acme-challenge` for domain verification, you don't need to worry about this. However, it is still recommended to have both IPv4 and IPv6 addresses allocated if your app is serving traffic.
```
andie787, line 58:
```suggestion
**Important:** Our hostname validation will fail without an IPv6 address and we won't attempt to issue or renew a certificate. If your app does not have one, you can allocate one with `flyctl ips allocate-v6`. If you use a CNAME `_acme-challenge` for domain verification, you don't need to worry about this. However, it is still recommended to have both IPv4 and IPv6 addresses allocated if your app is serving traffic.
```

andie787 (commented):
Thanks for adding this info. Left some tiny suggestions.

PR #1463, opened 2024-04-12
Add 'Volume placement' section to volume reference

IohannesArnold:
This started out as an effort to detail the meaning of the `--vm-xxx` flags, and grew into two new sections of the Volumes reference, as well as a rewrite of a few more lines, to clarify expectations and methods of volume placement.

### DIFF
--- a/reference/volumes.html.markerb
+++ b/reference/volumes.html.markerb
@@ -14,13 +14,17 @@ A Fly Volume is a slice of an NVMe drive on the physical server your Fly App run
 ## Volume considerations
 
 * **A volume belongs to one app**: Every Fly Volume belongs to a [Fly App](/docs/reference/apps/) and you can't share a volume between apps.
-* **A volume exists in one region**: A volume exists in only one region. Only a Machine running in the same region can access it.
+* **A volume exists on one server**: Each volume exists on one server in a single region. It is not network storage.
 * **A volume can attach to one Machine**: You need to run as many volumes as there are Machines. There's a one-to-one mapping between Machines and volumes. A Machine can only mount one volume at a time and a volume can be attached to only one Machine.
 * **Develop your app to handle replication**: Volumes are independent of one another; Fly.io does not automatically replicate data among the volumes on an app, so if you need the volumes to sync up, then your app has to make that happen.
 * **Create redundancy in your primary region**: If your app needs a volume to function, and the NVMe drive hosting your volume fails, then that instance of your app goes down. There's no way around that. You can run multiple Machines with volumes in your app's primary region to mitigate hardware failures.
 * **Create and store backups**: If you only have a single copy of your data on a single volume, and that drive fails, then the data is lost. Fly.io takes daily snapshots and retains them for 5 days, but the snapshots shouldn't be your primary backup method.
 
 Explore other options for data storage in [Databases & Storage](/docs/database-storage-guides/).
+
+## Volume attachment
+
+Fly Volumes and Fly Machines are meant to be paired together, but they are not necessarily always found in pairs. A Fly Volume can be created without a Fly Machine, or a Machine can be destroyed without destroying its volume. In these cases, the volume that’s left is called an "unattached" volume. A Fly Machine that does not require a volume will never attach itself to one. Fly Machines that do require volumes will *always* be attached to one, and any method of creating a new Fly Machine will always create a matching Fly Volume if there is not an unattached one available.
 
 ## Volume redundancy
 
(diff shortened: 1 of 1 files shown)


reference/volumes.html.markerb
```diff
@@ -30,11 +34,15 @@ Explore other options for data storage in [Databases & Storage](/docs/database-s
 
 In a few cases, you can run a single Machine with an attached volume. For example, if you're running an app with a standard SQLite database, or your app is in development and you're not yet worried about downtime.
 
-Many developers use volumes for databases, so if possible, each volume you create for your app is placed in a separate hardware zone by default. Note that having each volume in a separate hardware zone limits the number of volumes your app can have in a region. You can configure this setting with the `--require-unique-zone` option when you run [`fly volumes create`](/docs/flyctl/volumes-create/).
+## Volume placement
+
+You can create volumes at the same time you create Machines, or ahead of time. If you try to create new Machines through `flyctl scale count`, then `flyctl` will first check if you have any unattached volumes in the region you are creating new Machines in. If it finds unattached volumes, it will try to create the new Machines on the same servers. If it does not find any unattached volumes, then it will create new ones together with the Machines. You can also create unattached volumes through [`fly volumes create`](/docs/flyctl/volumes-create/).
```
andie787, line 39: This behaviour is described pretty well here: https://fly.io/docs/apps/scale-count/#scale-an-app-with-volumes.  It might be better to link to it if you think it's belongs in this doc?

reference/volumes.html.markerb
```diff
@@ -14,14 +14,18 @@ A Fly Volume is a slice of an NVMe drive on the physical server your Fly App run
 ## Volume considerations
 
 * **A volume belongs to one app**: Every Fly Volume belongs to a [Fly App](/docs/reference/apps/) and you can't share a volume between apps.
-* **A volume exists in one region**: A volume exists in only one region. Only a Machine running in the same region can access it.
+* **A volume exists on one server**: Each volume exists on one server in a single region. It is not network storage.
```
andie787, line 17: 💯

reference/volumes.html.markerb
andie787, line 39: To introduce volume placement, could we say something instead about how they can control placement by region, and in the case of GPUs by hardware, while we ensure that their Machines and volumes are on the same server and that volumes are spread across servers within a region.

reference/volumes.html.markerb
```diff
@@ -14,14 +14,18 @@ A Fly Volume is a slice of an NVMe drive on the physical server your Fly App run
...
 * **Create and store backups**: If you only have a single copy of your data on a single volume, and that drive fails, then the data is lost. Fly.io takes daily snapshots and retains them for 5 days, but the snapshots shouldn't be your primary backup method.
 
 Explore other options for data storage in [Databases & Storage](/docs/database-storage-guides/).
 
+## Volume attachment
+
+Fly Volumes and Fly Machines are meant to be paired together, but they are not necessarily always found in pairs. A Fly Volume can be created without a Fly Machine, or a Machine can be destroyed without destroying its volume. In these cases, the volume that’s left is called an "unattached" volume. A Fly Machine that does not require a volume will never attach itself to one. Fly Machines that do require volumes will *always* be attached to one, and any method of creating a new Fly Machine will always create a matching Fly Volume if there is not an unattached one available.
```
andie787, line 27: > Fly Machines that do require volumes will *always* be attached to one, and any method of creating a new Fly Machine will always create a matching Fly Volume if there is not an unattached one available.

The second part of this is only true for Fly Launch commands like `fly launch`, `fly deploy` and `fly scale count`; it doesn't work like that for the API or the `fly machine` commands, which I think will fail unless you create the volume first.

andie787 (commented):
I like the new headings! Left some suggestions/discussion topics to think about.

reference/volumes.html.markerb
```diff
@@ -14,14 +14,18 @@ A Fly Volume is a slice of an NVMe drive on the physical server your Fly App run
...
+
 ## Volume redundancy
```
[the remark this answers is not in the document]
IohannesArnold: I think that's something that should be said, but it should go in some place about scaling & availability in general, not Volumes specifically.

Actually, I'm not sure how much about redundancy needs to be in this Volumes reference at all. I get the sense that this was written in the Nomad era, when Allocs floated between hosts while Volumes, distinctly, did not. Now that Machines are just as pinned to specific servers as Volumes, it would be better to have a general document about scaling, redundancy, and availability, rather than trying to pack it into Volumes.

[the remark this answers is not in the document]
andie787: We have this (hasn't been updated for a bit though): https://fly.io/docs/reference/app-availability/

I think we want to keep mentioning redundancy when we talk about volumes because of the risk of single host failures, right?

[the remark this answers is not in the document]
IohannesArnold: Update to include what you'd said.

andie787:
@IohannesArnold Feel free to merge this now if you're happy with it!

PR #1466, opened 2024-04-15
Document proxy error codes

jsierles:
This PR documents the proxy error codes introduced in https://github.com/superfly/fly-proxy/pull/1203.


reference/error-codes.html.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Fly.io Error Codes
+layout: docs
+sitemap: false
+nav: firecracker
+---
+
+The Fly.io platform logs internal errors from its various parts such as the proxy, VM orchestrator, etc. Each log line contains an error code, and a field containing that code. This page gives more context about errors.
```
andie787, line 8: We should say "Fly Proxy", not sure what the "VM orchestrator" is?

PR #1468, opened 2024-04-15
Document new 'http_options.response.pristine' flag

pborzenkov:


andie787 (commented):
Thanks for adding this!

Could you add a sentence or two about why this would be useful? This will help when we update the docs.

pborzenkov:
> Thanks for adding this!
> 
> Could you add a sentence or two about why this would be useful? This will help when we update the docs.

Done!

PR #1476, opened 2024-04-18
added troubleshooting tips to private networking doc

brainlid:
### Summary of changes
Add some troubleshooting tips to the Private Networking VPN docs for `fly wireguard`.

### Preview
image
image

### Notes

I had some connectivity issues from my home machine to a Fly hosted app and I struggled to narrow down where the issue was. When I got down to the equivalent of "Is it plugged in?", I realized I didn't know how to troubleshoot my Fly WireGuard connections.

In the end, I had a WireGuard connection open to a different org than where the app was running. I _thought_ the app was in the other org. I believe this will be a more common problem for people working with both personal and business orgs.

This attempts to help people solve this type of problem more easily on their own.


networking/private-networking.html.md
```diff
@@ -298,3 +298,61 @@ To list all the tunnels set up for an organization, run `fly wireguard list`. Yo
...
+
+If results are returned, you have a VPN connection to an org at Fly.io and the results list all the app names.
+
+If no results are returned, either you do not have a WireGuard VPN connection open or there are NO apps running in the org.
+
+#### Am I connected to the right Fly.io org?
+
+WireGuard connections are created to a specific org. Each org's network is isolated from other orgs. You may have a VPN connection to Fly.io, but it may be to a different org than where your app is located. By default, you are working with your **personal** organization. If the app is in a business focused org or a shared org, you need to ensure your VPN connection is the org where the app lives.
+
+We can also try and ping our app's machine.
```
andie787, line 325:
```suggestion
You can also try to ping your app's Machine.
```

networking/private-networking.html.md
```diff
@@ -298,3 +298,61 @@ To list all the tunnels set up for an organization, run `fly wireguard list`. Yo
 #### Remove a tunnel
 
 To remove a tunnel, run `fly wireguard remove`. You can specify the organization and tunnel name on the command line or be prompted for both.
+
+### Troubleshooting
+
+Having trouble connecting to a Fly.io hosted resource? The Private Network VPN feature makes it easy for a local machine to establish a fast, reliable connection to a Fly.io organization. However, when you can't connect to something, it's helpful to establish a baseline of what is, or what is not, working.
```
andie787, line 304: Suggest re-wording this slightly. Maybe just remove the second sentence and the "however" from the third sentence? We don't need to sell the feature here and we'll usually try to avoid saying something is "easy" (at least in straight-up tech docs and especially in troubleshooting docs).

andie787 (commented):
This is a great idea! 
Left some suggestions. You're good to merge after that!

PR #1480, opened 2024-04-22
add ref docs on autoscaling

senyosimpson:
### Summary of changes

Adds reference documentation on autoscaling capabilities on our platform

### Notes

I'm not sure how in depth the reference should be. Both link out to the respective features and documentation, just providing a short overview of the features. Also, the name "pre-allocated machine autoscaling" sucks but i'm not sure how else to describe it. Open to ideas!

### DIFF
--- a/reference/autoscaling.html.md
+++ b/reference/autoscaling.html.md
@@ -18,7 +18,8 @@ For a comprehensive overview of how it works, see the [documentation](https://fl
 
 ## Metrics-based autoscaling
 
-This feature is supported through [Fly Autoscaler](https://github.com/superfly/fly-autoscaler) which you deploy as an app into your organisation.
-It polls metrics from a Prometheus instance and computes the number of machines needed based on those metrics. The autoscaler will create and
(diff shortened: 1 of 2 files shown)


reference/autoscaling.html.md
```diff
@@ -0,0 +1,25 @@
...
+sitemap: false
+nav: firecracker
+---
+
+Autoscaling is used to increase the number of Fly Machines based on demand. There are two forms of autoscaling supported:
+
+1. Pre-allocated Machine autoscaling
+2. Metrics-based autoscaling
+
+## Pre-allocated Machine autoscaling
```
andie787, line 13: This title is okay...? We usually just call it "auto start and stop", which is also kind of an awkward feature name. 

What about "Traffic-based Machine shutdown and startup"? Or something similar?

senyosimpson: i think the places where we describe it as `auto start and stop`, it sounds fine because its context of something (i.e configuration options, etc). I think the title as is is good enough, though i don't like it very much

PR #1495, opened 2024-04-29
use a gpu app in gpu example

senyosimpson:
### Summary of changes

Use an actually relevant example


andie787 (approved):

andie787:
@senyo merge when ready. Sorry for the delay!

PR #1510, opened 2024-05-06
Update custom-domains-with-fly.html.md

vorant94:
### Summary of changes

As `fly certs --help` shows there is no more `fly certs create` command. Based on the commands that are there, I assume it was replaces with `fly certs add`, so I suppose this particular documentation page is out of date, hence here is the fix

the output of `fly certs --help`:
```cmd
Manages the certificates associated with a deployed application. Certificates are created by associating a
hostname/domain with the application. When Fly is then able to validate that hostname/domain, the platform gets
certificates issued for the hostname/domain by Let's Encrypt.

Usage:
  fly certs [command]

Available Commands:
  add         Add a certificate for an app.
  check       Checks DNS configuration
  list        List certificates for an app.
  remove      Removes a certificate from an app
  show        Shows certificate information

Flags:
  -h, --help   help for certs

Global Flags:
  -t, --access-token string   Fly API Access Token
      --debug                 Print additional logs and traces
      --verbose               Verbose output

Use "fly certs [command] --help" for more information about a command.

```

### DIFF
--- a/networking/custom-domains-with-fly.html.md
+++ b/networking/custom-domains-with-fly.html.md
@@ -63,13 +63,13 @@ Create an A record pointing to your IPv4 address, and an AAAA record pointing to
 Once these settings are in place, you can add the custom domain to the application's certificates. If we are configuring example.com, then we would simply run:
 
 ```cmd
-fly certs create example.com
+fly certs add example.com
 ```
 
 If you need a wildcard domain, remember to place quotes around the hostname to avoid shell expansion:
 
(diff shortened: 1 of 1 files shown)


vorant94:
Any update on it?

andie787:
thanks for this update @vorant94. It makes sense to use the `add` command in the docs, however, you will notice if you try running the `fly certs create` command, that it still works as expected. Maybe we should document it as an alias!

PR #1512, opened 2024-05-06
Add supported tlds

Lucais11:
### Summary of changes
Added some information about the top level domains we currently support. This should help customers who use custom domains.
### Preview

Screenshot 2024-05-06 at 1 34 15 PM

### DIFF
--- a/networking/custom-domain.html.markerb
+++ b/networking/custom-domain.html.markerb
@@ -86,6 +86,9 @@ those errors aren't ever visible, you can use a DNS challenge to pre-generate th
 To do that, you need to create a `CNAME` DNS record for a subdomain, `_acme-challenge`, of your domain (the DNS Validation host name)
 and point it at the DNS Validation Target. The process will depend on your DNS provider. Once complete, and the updated DNS data has propagated, that domain will be queried and confirm you have control of it. Certificates will be generated and installed and you'll be able to access your custom domain.
 
+## Supported top-level domains
+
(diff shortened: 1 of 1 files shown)


networking/custom-domain.html.markerb
```diff
@@ -86,6 +86,9 @@ those errors aren't ever visible, you can use a DNS challenge to pre-generate th
 To do that, you need to create a `CNAME` DNS record for a subdomain, `_acme-challenge`, of your domain (the DNS Validation host name)
 and point it at the DNS Validation Target. The process will depend on your DNS provider. Once complete, and the updated DNS data has propagated, that domain will be queried and confirm you have control of it. Certificates will be generated and installed and you'll be able to access your custom domain.
 
+## Supported top-level domains
+
+Here is the list of custom Top-Level Domains (TLDs) we currently support: [Supported TLDs](https://data.iana.org/TLD/tlds-alpha-by-domain.txt). Please note that we do not regularly update this list. It reflects the TLDs we support as of today, and there may be delays in including the newest ones.
```
andie787, line 91: The reason I'm suggesting a few changes here is just to make it clear where the link will take them and to replace "today" with a more general statement:
 ```suggestion
 We support the top-level domains on the [IANA list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt+external). Note that we only periodically update top-level domain support and there might be a delay before we add support for new top-level domains.
```
Or something like that? 
(added the +external to the end of the URL, it adds a little icon so users know they're being sent out of Fly docs)

Lucais11: Oh yeah this works! I'll commit this suggestion

PR #1517, opened 2024-05-09
Update the Next.js guide

mentels:
### Summary of changes

Generate the Next.js app with `create-next-app`

### Preview
image
image
image
image

n/a

### Notes

The update is about generating a new Next.js app with `create-next-app` instead of redirecting the user to a tutorial at nextjs.org

The're are other suspects for improvement in the guide:
- I haven't noticed any improvement in the image size after applying the "standalone" config option
- ~~the `COPY` instructions suggested as an alternative make the resulting VM crash with~~

But I'll leave it for later / others since I'm just wrapping my head around the Next.js.


rubys (approved):

jeregrine (approved):
Fix the vale notes and LGTM

PR #1534, opened 2024-05-20
Restructure Private Networking page

IohannesArnold:
* Move the table with `.internal` subdomains to the front as it is the most relevant
 * Add new subsection about the details of 6PN addresses, with a warning that 6PN addresses are not stable.
 * Make some other small tweaks to improve flow.


networking/private-networking.html.md
```diff
@@ -150,7 +131,7 @@ VERSION	IP                	TYPE   	REGION	CREATED AT
 v6     	fdaa:0:22b7:0:1::3	private	global	just now
 ```
 
-If you want to expose services to another Fly.io organization you have access to, then use the `--org` flag.
+You can also use Flycast to expose an app in one Fly organization to another Fly organization. If you want to expose services to another Fly.io organization you have access to, then use the `--org` flag.
```
andie787, line 134:
```suggestion
You can also use Flycast to expose an app in one Fly organization to another Fly organization by using the `--org` option when you allocate the Flycast address:
```

networking/private-networking.html.md
```diff
@@ -10,115 +10,95 @@ redirect_from:
...
 
-Since this is the default configuration we set up for Machines on Fly.io, you probably don't need to do anything special to make this work; if your app shares an organization with another app called `random-potato-45`, then you should be able to `ping6 random-potato-45.internal`.
+Underneath `.internal` there are second-level domains for every app in your Fly organization. For example, if your app is in an organization with another app called `random-potato-45`, then there will be a AAAA record at `random-potato-45.internal`. The AAAA record will contain *all* the 6PN addresses of the started Fly Machines that make up the `random-potato-45` Fly App. Note that different libraries and tools will use multi-address AAAA records differently; most will only use the first address that is returned, but others might round-robin between entries for every request -- if you'd like to know more, consult the documentation for the library or tool you are using for DNS lookup.
```
andie787, line 23: Curious why we'd use `random-potato-45` instead of `my-app-name` (since we're tending to standardize this usage throughout the docs)?

andie787 (commented):
Left some suggestions and questions for you

I think the org changes are a big improvement!

networking/private-networking.html.md
```diff
@@ -10,115 +10,95 @@ redirect_from:
...
-Address = fdaa:0:18:a7b:d6b:0:a:2/120
-DNS = fdaa:0:18::3
-```
+Examples of retrieving this information are in the [fly-examples/privatenet](https://github.com/fly-apps/privatenet) repository.
 
-You guessed it; it's the `DNS` line. Your DNS server address will also start with `fdaa:`, but the next two parts are unique to your organization's network. All 6PN addresses are prefixed by the organization's network ID; that's the part of the address that locks it to your organization.
+## Listening for connections on 6PN addresses
 
-All our WireGuard DNS addresses follow this pattern: take the organization prefix, and tack `::3` onto the end. In this example, the Wireguard peer address is:
+When deploying a Fly Machine, we append to the Machine's `/etc/hosts` to alias the 6PN address of the Machine to `fly-local-6pn`.
```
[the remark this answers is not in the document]
IohannesArnold: Not on product so I'm not certain, but I believe this is only once when creating the rootfs.

networking/private-networking.html.md
IohannesArnold: Ha, not my idea! It was already there as an example, so for consistency I extended its use throughout the page. But I didn't know there was an even higher standard for all the docs. I'll switch to `my-app-name`.

PR #1543, opened 2024-05-27
Add more CI/CD resources

almirsarajcic:


andie787:
hi @almirsarajcic 

we're not able to test this at the moment and don't generally accept docs contributions that are self-promoting

the community forum can be a good place to discuss specific solutions!

almirsarajcic:
I saw the Felt example above so I thought I too could use this opportunity to increase the visibility of a good solution.

But it's ok, I'm not going to push it further.

PR #1558, opened 2024-06-06
Replace Redis example output with Supabase output in Supabase docs

spencerchristensen:
Update `fly ext supabase status` to use Supabase and not Redis output. The demo `late-waterfall-1133` name appears to have been lifted from `redis.html.md`.


andie787:
oh good catch! Thanks!

PR #1582, opened 2024-06-17
Arcjet extension docs

davidmytton:
### Summary of changes

Adds the [Arcjet](https://arcjet.com/) extension documentation.

### Notes

cc @jsierles


reference/arcjet.html.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Get started
+
+To configure your application to use Arcjet, run this command from your Fly.io application directory:
+
+```cmd
+flyctl ext arcjet create
+```
+
+Then set the `ARCJET_KEY` secret on your app.
```
jsierles, line 20: If you run the command in your app directory, the key will be set automatically. If you set it outside the app directory, you'd need to manually run this command.

jsierles (changes requested):

jsierles (approved):
Generally looks good. It would be good to add a section about pricing, even if it isn't worked out yet, and also a callout about being in beta, and what that means. Check the Supabase docs for details on that: https://fly.io/docs/reference/supabase/

jsierles (approved):

davidmytton:
@jsierles looks like the Vale linter needs updating to allow lower-case `arcjet` in the URLs.

jsierles:
You can add here https://github.com/superfly/docs/blob/main/styles/config/vocabularies/fly-terms/accept.txt

PR #1583, opened 2024-06-17
Add docs for fly-autoscaler

andie787:
### Summary of changes

Add docs for fly-autoscaler, the metrics-based autoscaler app.

### DIFF
--- a/apps/autoscale-by-metric.html.markerb
+++ b/apps/autoscale-by-metric.html.markerb
@@ -84,7 +84,7 @@ Metrics collection settings:
 Autoscaling settings:
 
 - `FAS_APP_NAME` is the name of the target application to scale.
-- `FAS_CREATED_MACHINE_COUNT` defines an [Expr](https://expr-lang.org/) expression to calculate the number of Machines to create.
+- `FAS_CREATED_MACHINE_COUNT` defines an [Expr](https://expr-lang.org/) expression to calculate the number of Machines to create or destroy.
 
 This example expression assumes that each Machine could handle two items in the queue and uses the `min()` function to prevent the autoscaler from scaling more than `50` Machines.
 
--- a/apps/autoscale-by-metric.html.markerb
+++ b/apps/autoscale-by-metric.html.markerb
@@ -1,0 +1,181 @@
+---
+title: Autoscale based on metrics
+layout: docs
+nav: firecracker
+---
+
+Most applications on Fly.io can use the request-based autoscaling that is built into our proxy by setting [the soft & hard limits in your fly.toml](/docs/reference/configuration/#http_service-concurrency). However, some applications need to scale on other metrics besides requests such as a work queue depth. In these instances, you can use our open source [fly-autoscaler](https://github.com/superfly/fly-autoscaler) to scale your workload.
+
+The metrics-based autoscaler works by collecting metrics from different sources, such as Prometheus or Temporal, and computing the number of required Machines based on those metrics. This reconciliation process happens on a loop every 15 seconds by default. We use the [Expr language](https://expr-lang.org/) for defining target Machine counts which gives a rich set of built-in arithmetic functions.
+
+## Quickstart
+
+The `fly-autoscaler` is an application that you can run and customize yourself to work with your specific scaling needs. To get up and running, you'll set up the app, configure secrets, set the configuration, and deploy the autoscaler.
+
+As a prerequisite, you need an existing target application that you want to scale and a user-defined metric to scale on. In this example, you'll scale on a metric called `queue_depth` but you can replace that with your own.
+
+### Create the autoscaler application
+
+First, create a new Fly.io application that will run the autoscaler. This application will run within your organization so you have full control over it. Replace the `my-autoscaler` name with a unique name for your autoscaler application.
+
+```
+fly apps create my-autoscaler
+```
+
+### Acquire a deploy token
+
+The first auth token you'll need is one that has permissions to deploy your target app:
+
+```cmd
+fly tokens create deploy -a my-target-app
+```
+
+Copy the resulting token and set it as a secret on your autoscaler app:
+
+```cmd
+fly secrets set -o my-autoscaler --stage FAS_API_TOKEN="FlyV1 ..."
+```
+
+## Acquire a token to read from Prometheus
+
+The next auth token you'll need is one that has permissions to read from your organization's Prometheus data on Fly:
+
+```cmd
+fly tokens create readonly -o my-org
+```
+
+Copy the token and use it as a secret on your autoscaler app:
+
+```cmd
+fly secrets set -o my-autoscaler --stage FAS_PROMETHEUS_TOKEN="FlyV1 ..."
+```
+
+### Configure your autoscaler fly.toml
+
+Next, set up a `fly.toml` configuration file for your autoscaler to set a few environment variables. Replace the `my-autoscaler`, `my-target-app`, and `my-org` with values for your situation.
+
+```toml
+app = "my-autoscaler"
+
+[build]
+image = "flyio/fly-autoscaler:0.3"
+
+[env]
+FAS_PROMETHEUS_ADDRESS = "https://api.fly.io/prometheus/my-org"
+FAS_PROMETHEUS_METRIC_NAME = "qdepth"
+FAS_PROMETHEUS_QUERY = "sum(queue_depth{app='$APP_NAME'})"
+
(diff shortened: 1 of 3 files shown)


apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,181 @@
+---
+title: Autoscale based on metrics
+layout: docs
+nav: firecracker
+---
+
+Most applications on Fly.io can use the request-based autoscaling that is built into our proxy by setting [the soft & hard limits in your fly.toml](/docs/reference/configuration/#http_service-concurrency). However, some applications need to scale on other metrics besides requests such as a work queue depth. In these instances, you can use our open source [fly-autoscaler](https://github.com/superfly/fly-autoscaler) to scale your workload.
```
andie787, line 7: @benbjohnson Suggest starting with what it is/does and what apps it's good for. Then say that most apps can use auto start and stop. Something like this but with your better understanding/edits:

```suggestion
The metrics-based autoscaler scales an app's Machines based on any metric, such as pending work items or queue depth. Scaling on metrics other than requests is useful for apps that aren't running web services, like background workers, for example. Apps with services that need to scale based on HTTP requests can use the [auto stop and start feature](https://fly.io/docs/apps/autostart-stop/) for Machines that's built into the Fly Proxy.
```

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,181 @@
...
+**Note:** The autoscaler creates new Machines in an application by cloning existing ones so it is not able to scale to zero. It will always keep at least one Machine running.
+</div>
+
+You can find a full working example of the autoscaler at our [fly-autoscaler-example](https://github.com/fly-apps/fly-autoscaler-example) repo.
+
+## Other use cases
+
+### Start and stop Machines
+
+If you already have a pool of created Machines that you wish to autoscale, you can use the `FAS_STARTED_MACHINE_COUNT` expression instead of creating and destroying Machines.
```
andie787, line 111: @benbjohnson Would this be better?
```suggestion
If you already have a pool of created Machines that you want to autoscale, you can use the `FAS_STARTED_MACHINE_COUNT` expression instead of creating and destroying Machines with `FAS_CREATED_MACHINE_COUNT`.
```

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,181 @@
+---
+title: Autoscale based on metrics
+layout: docs
+nav: firecracker
+---
+
+Most applications on Fly.io can use the request-based autoscaling that is built into our proxy by setting [the soft & hard limits in your fly.toml](/docs/reference/configuration/#http_service-concurrency). However, some applications need to scale on other metrics besides requests such as a work queue depth. In these instances, you can use our open source [fly-autoscaler](https://github.com/superfly/fly-autoscaler) to scale your workload.
+
+The metrics-based autoscaler works by collecting metrics from different sources, such as Prometheus or Temporal, and computing the number of required Machines based on those metrics. This reconciliation process happens on a loop every 15 seconds by default. We use the [Expr language](https://expr-lang.org/) for defining target Machine counts which gives a rich set of built-in arithmetic functions.
```
andie787, line 9: @benbjohnson 
Should we also explicitly say something like:

> Your target number of Machines can be a single number or a range. Using a range for scaling can protect against too many Machines being started/stopped or created/destroyed in a short time period when you're using a more volatile metric.

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,181 @@
...
+
```
andie787, line 10: @benbjohnson Do you think it would be useful to talk about create/destroy vs. stop/start here? Or down below in the Stop/start Machines section?
Something like:
```suggestion
You have the option of scaling to a target by starting and stopping a group of existing Machines, or by creating and destroying Machines. There are benefits and drawbacks to each of these options.

If you scale by starting and stopping existing Machines, then your Machines will start up very quickly. You will pay for the Machines's CPU and RAM when they're running and for the rootfs when they're stopped.

If you scale by creating and destroying Machines, then your Machines will be slightly slower to reach a `started` state since it takes longer to cerate a Machine than to start one. You won't need to create a "pool" of Machines. You will only pay for the Machines's CPU and RAM when they're running and won't need to pay for rootfs since the Machines are destroyed when not needed.
```

andie787:
@benbjohnson Made some smaller fixes from the original draft but left comments for you about bigger changes.

benbjohnson (approved):
All your suggestions look good to me. 👍

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,209 @@
...
+The quickstart describes how to configure the autoscaler with environment variables. You can also configure the autoscaler with a YAML config file if you don't want to use environment variables or if you want to configure more than one metric collector.
+
+See the reference [fly-autoscaler.yml](https://github.com/superfly/fly-autoscaler/blob/main/etc/fly-autoscaler.yml) file for an example and more details.
+
+### Autoscaler config
+
+- `FAS_APP_NAME`: The name of the target app to scale.
+- `FAS_CREATED_MACHINE_COUNT`: An [Expr](https://expr-lang.org/) expression to calculate the number of Machines to create or destroy.
+- `FAS_STARTED_MACHINE_COUNT`: An [Expr](https://expr-lang.org/) expression to calculate the number of Machines to stop or start.
```
andie787, line 194: I see that there are still some `MIN` and `MAX` versions of `FAS_CREATED_MACHINE_COUNT` and `FAS_STARTED_MACHINE_COUNT`. Do we include these in the reference or do we just use the Expr expression to calculate these values internally?

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,209 @@
...
+
+### Prometheus collector
+
+- `FAS_PROMETHEUS_ADDRESS`: The URL of the Prometheus endpoint to query.
+- `FAS_PROMETHEUS_METRIC_NAME`: The local variable name that the metric result will be stored as.
+- `FAS_PROMETHEUS_QUERY`: The Prometheus query to run.
+
+
+### Temporal collector
```
andie787, line 203: I see a `FAS_TEMPORAL_QUERY` variable, but it's not in the Fresh Produce or this doc? What is it and should we include it?

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,209 @@
...
+
+See the reference [fly-autoscaler.yml](https://github.com/superfly/fly-autoscaler/blob/main/etc/fly-autoscaler.yml) file for an example and more details.
+
+### Autoscaler config
+
+- `FAS_APP_NAME`: The name of the target app to scale.
+- `FAS_CREATED_MACHINE_COUNT`: An [Expr](https://expr-lang.org/) expression to calculate the number of Machines to create or destroy.
+- `FAS_STARTED_MACHINE_COUNT`: An [Expr](https://expr-lang.org/) expression to calculate the number of Machines to stop or start.
+
+### Prometheus collector
```
andie787, line 196: I see a `FAS_PROMETHEUS_TOKEN` variable too. What is it and should we include it?

apps/autoscale-by-metric.html.markerb
```diff
@@ -0,0 +1,209 @@
+---
+title: Autoscale based on metrics
+layout: docs
+nav: firecracker
+---
+
+The metrics-based autoscaler scales an app's Machines based on any metric, such as pending work items or queue depth. Scaling on metrics other than requests or connections is useful for apps, like background workers, that aren't running web services. Apps with services that need to scale based on HTTP requests can use the built-in Fly Proxy [autostop/autostart feature](https://fly.io/docs/apps/autostart-stop/) for Machines.
+
+The autoscaler works by collecting metrics from different sources, such as Prometheus or Temporal, and computing the number of required Machines based on those metrics. This reconciliation process happens on a loop every 15 seconds by default. We use the [Expr language](https://expr-lang.org/) for defining target Machine counts, which gives a rich set of built-in arithmetic functions.
```
andie787, line 9: Should we say how you can change the reconciliation interval in the yml config? Can you use a combination of yaml config and env variables?

andie787 (commented):
@benbjohnson 
Final final review!

andie787:
@benbjohnson I'm going to publish as-is, since it's fairly complete, and we can revisit my questions when there's time.

PR #1589, opened 2024-06-18
Fix link to extension references in the pricing page

gargakshit:
### Summary of changes

Links to tigris, supabase and upstash references were broken in the pricing page.


jsierles:
good catch!

PR #1630, opened 2024-07-01
Publish Machine API rate limits

kylemclaren:
Screenshot 2024-07-01 at 17 08 16

https://community.fly.io/t/understanding-fly-io-api-rate-hard-limits/20636

### DIFF
--- a/machines/api/working-with-machines-api.html.markerb
+++ b/machines/api/working-with-machines-api.html.markerb
@@ -8,13 +8,13 @@ order: 1
 
 The Fly Machines API provides resources to create, destroy, and manage Fly Apps, Fly Machines, and Fly Volumes. Refer to the following reference docs for the usage of each:
 
-**[Apps resource](/docs/machines/api/apps-resource/)** — Every Fly Machine belongs to a [Fly App](/docs/reference/apps/).
+**[Apps resource](/docs/machines/api/apps-resource/)** - Every Fly Machine belongs to a [Fly App](/docs/reference/apps/).
 
-**[Machines resource](/docs/machines/api/machines-resource/)** — Fly Machines are our fast-launching VMs.
+**[Machines resource](/docs/machines/api/machines-resource/)** - Fly Machines are our fast-launching VMs.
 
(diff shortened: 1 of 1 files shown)


machines/api/working-with-machines-api.html.markerb
```diff
@@ -73,4 +73,10 @@ Typically, 2xx HTTP status codes denote successful operations, 4xx codes imply f
 | 401 | The API key used was missing or invalid. |
 | 404 | The resource was not found. |
 | 408 | Request timed out. |
-| 5xx | Indicates an error with Fly.io API servers. |
\ No newline at end of file
+| 5xx | Indicates an error with Fly.io API servers. |
+
+## Rate Limits
+
+Machine API rate limits apply _per-action_, _per-machine_. Currently, the limit is 1 request, per second, per action (i.e. Create Machine, Start Machine etc.)—short-term burstable up to 3 req/s, per action.
```
andie787, line 80: Suggest removing "Currently" since it might imply an incoming change.

andie787, line 80: Suggest using "burst limit" in place of "burstable" here and below. I feel like burst limit pairs with rate limit and is a familiar term. What do you think?

andie787 (commented):
Thank you for adding this missing info!!

Left a few suggestions. When you're happy with it, go ahead and merge.

catflydotio (commented):
The content change looks great to me; I wouldn't make the change to the dashes earlier in the document, the argument being that that's effectively a style/formatting policy; i.e. a site-wide decision. We shouldn't change those in individual updates.

kylemclaren:
@catflydotio Gotcha, how about removing the one em dash?

image

andie787:
> The content change looks great to me; I wouldn't make the change to the dashes earlier in the document, the argument being that that's effectively a style/formatting policy; i.e. a site-wide decision. We shouldn't change those in individual updates.

No problem with the word. It's more about the context. Since we're sharing AWS links 😆 

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-request-throttling.html

kylemclaren:
Actually, I think you are right

PR #1638, opened 2024-07-02
Discontinued Hobby plans clarification

andie787:
### Summary of changes

- the paid hobby plan is deprecated
- no new users can sign up for any Hobby plans
- if you change from the hobby plan, you can't go back
- the bit about no free allowances during the free trial was already there, but moved it to the top of the free allowances section


asib (approved):
Looks great. Only suggestion would be to prefix the plan with `(deprecated)` as opposed to suffixing, but that's more just personal preference :)

jphenow (approved):
:shipit:

andie787:
> Looks great. Only suggestion would be to prefix the plan with `(deprecated)` as opposed to suffixing, but that's more just personal preference :)

I tried this out and it made sense and looks better in a few places, but I worry that it doesn't work in the list of plans (it might imply they are all deprecated) so I left it there.

PR #1643, opened 2024-07-04
Add "Autoscale machines like a boss" blueprint

pborzenkov:


andie787 (commented):
This is good to go if you're happy with it @pborzenkov!

I might come back and make a few minor edits when we settle on the naming of and/or rename the Fly Proxy autostart/autostop.

PR #1644, opened 2024-07-04
Improve FAS_CREATED_MACHINE_COUNT description in autoscale-by-metric.…

MaxLap:
"expression to calculate the number of Machines to create or destroy." Makes it sound like if I return 1, it's to create 1 machine.

But the value returned is instead the target to to scale up/down to.

### Summary of changes

`expression to calculate the number of Machines to create or destroy.`

Becomes

`expression to calculate the target number of Machines, scaling via either create or destroy.`


andie787:
This is a very good point. Took your advice to update the descriptions but created a new PR since we're going with slightly different wording. #1713 

Thanks for bringing this up!

Issue #1578, opened 2024-06-14
"Running Fly Apps On UDP and TCP" doc does not mention dedicated IP requirement


cixel:
Title: Running Fly Apps On UDP and TCP
Location: https://fly.io/docs/networking/udp-and-tcp/
Source: https://github.com/superfly/docs/blob/main/networking/udp-and-tcp.html.markerb

### Describe the issue

What the title says.

I had trouble standing up [headscale](https://github.com/juanfont/headscale)'s embedded DERP server, and eventually realized that the issue was that UDP packets won't make it through to the application over shared IPv4.

This is documented in the [Public Network Services](https://fly.io/docs/networking/services/#dedicated-ipv4) doc, but not mentioned anywhere in the doc for setting up UDP. Given that UDP is also unsupported over dedicated IPv6, this means that a paid/dedicated IPv4 address is a requirement for UDP, which seems like a pretty meaningful "gotcha" and should be documented.

The documentation is otherwise quite thorough and even includes an example/reproducer, which is great. I suspect it was just written before the change to default to shares IPs for IPv4.

andie787:
fixed by #1657

PR #1657, opened 2024-07-08
Update the UDP doc a little

andie787:
### Summary of changes

- simplified
- moved the configuration/gotchas to the top
- stopped calling them gotchas and removed multiple "we're working on it"s
- reworded the headings and moved them up a level
- seemingly random edits
- left in the bit about eBPF magic despite my instinct to edit that too (might go back to it though)


andie787:
Fixed #1578

PR #1659, opened 2024-07-09
Include a Tigris section for the go docs page

KTanAug21:
### Summary of changes
Add a section in the Go docs page that includes quick steps on creating and connecting with a Tigiris Bucket

### DIFF
--- a/languages-and-frameworks/golang.html.markerb
+++ b/languages-and-frameworks/golang.html.markerb
@@ -6,6 +6,12 @@ redirect_from: /docs/getting-started/golang/
 
 <%= partial "partials/intro", locals: { runtime: "Go" } %>
 
+<div class="note icon">
+Already have a Go Fly App? Integrate it with [Tigris for file storage!](https://www.tigrisdata.com/docs/concepts/architecture/)
+
+It's simple, it's scalable, it's secure&mdash;it's your bucket in the cloud with effortless CDN-magic. It's easy to integrate with Go, [try it out](docs/languages-and-frameworks/golang/#add-tigris-global-storage-optional).
+</div>
+
 ## _The Example Application_
 
 You can get the code for the example from [the GitHub repository](https://github.com/fly-apps/go-example). Just `git clone https://github.com/fly-apps/go-example` to get a local copy.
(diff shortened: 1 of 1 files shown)


languages-and-frameworks/golang.html.markerb
```diff
@@ -204,6 +204,99 @@ opening https://golang-example.fly.dev ...
 
 Your browser will be sent to the displayed URL.
 
+## _Tigris Global Storage( Optional )_
```
andie787, line 207: Suggest:
```suggestion
## _Add Tigris Global Storage (Optional)_
```

languages-and-frameworks/golang.html.markerb
```diff
@@ -204,6 +204,99 @@ opening https://golang-example.fly.dev ...
...
+	Setting the following secrets on ...:
+	AWS_ACCESS_KEY_ID: <AWS_ACCESS_KEY_ID_VALUE>
+	AWS_ENDPOINT_URL_S3: https://fly.storage.tigris.dev
+	AWS_REGION: auto
+	AWS_SECRET_ACCESS_KEY: <AWS_SECRET_ACCESS_KEY>
+	BUCKET_NAME: <BUCKET_NAME>
+	```
+	
+	Of course, this would only automatically happen if the create command was run in a directory containing a `fly.toml` for our Go Fly app. 
+	On the otherhand, if that wasn't the case, we can still manually set necessary [secrets](https://fly.io/docs/flyctl/secrets/) likeso:
```
andie787, line 267:
```suggestion
	But if that wasn't the case, we can still manually set necessary [secrets](https://fly.io/docs/flyctl/secrets/) likeso:
```

languages-and-frameworks/golang.html.markerb
```diff
@@ -204,6 +204,99 @@ opening https://golang-example.fly.dev ...
...
+	```cmd
+	fly secrets set AWS_ACCESS_KEY_ID="<TIGRIS_ACCESS_KEY_ID>"
+	fly secrets set AWS_SECRET_ACCESS_KEY="<TIGRIS_SECRET_ACCESS_KEY>"
+	```
+
+5. <b>Upload a file:</b> Aside from connecting with a Bucket, we can use other [AWS SDK functions](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html). Let's use PutObject function to  upload a text file on the go likeso:
```
andie787, line 273:
```suggestion
5. <b>Upload a file:</b> Aside from connecting with a Bucket, we can use other [AWS SDK functions](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html). Let's use the `PutObject` function to  upload a text file on the go like so:
```

andie787 (commented):
Nice! I left a couple small suggestions. Also, what do you think of calling this section out at the top of the page after the intro partial? Something like "Already have a Go app running? Check out the section on adding "Tigris Object Storage". Or something better than that :)

KTanAug21:
> Nice! I left a couple small suggestions. Also, what do you think of calling this section out at the top of the page after the intro partial? Something like "Already have a Go app running? Check out the section on adding "Tigris Object Storage". Or something better than that :)
> 
> PS I think this is a fine place for this right now since we don't have have a go section. And you should still link to it from https://fly.io/docs/reference/tigris/#use-tigris-with-your-framework.

Thanks so much for the review @andie787 !

PR #1665, opened 2024-07-10
Fix links in autostart-internal-apps blueprint

matttpt:
### Summary of changes

A small fix for two links.

### Notes


andie787:
😱 thanks!

PR #1667, opened 2024-07-11
Added a wordpress setup page

Johannes-Werbrouck:


languages-and-frameworks/wordpress.html.markerb
```diff
@@ -0,0 +1,160 @@
+---
+title: Run a Wordpress app
```
andie787, line 2: Use WordPress here and throughout since that's how they style the word.

Also capitalize "App" (trying to move to sentence case headings but the frameworks section uses the heading for the sidenav so we need to be consistent with the rest)

languages-and-frameworks/wordpress.html.markerb
```diff
@@ -0,0 +1,160 @@
...
+layout: language-and-framework-docs
+redirect_from: /docs/getting-started/wordpress/
```
andie787, line 4: you can remove this redirect

languages-and-frameworks/wordpress.html.markerb
```diff
@@ -0,0 +1,160 @@
...
+App Machines: shared-cpu-1x, 1GB RAM (most apps need about 1GB of RAM)
+Postgres:     <none>                 (not requested)
+Redis:        <none>                 (not requested)
+
+? Do you want to tweak these settings before proceeding? Yes
+```
+
+Type `n` if you're happy with the defaults listed, otherwise type `y` to open the Fly Launch web page and edit your settings, including:
+
+* Name: Keep the default app name or enter your own.
```
andie787, line 37: Maybe add something like, The name will be part of URL for your app, unless you use a custom domain.

languages-and-frameworks/wordpress.html.markerb
```diff
@@ -0,0 +1,160 @@
...
+
+* Region: Keep the fastest [region](/docs/reference/regions/) (the value of `primary_region` in the `fly.toml` file) as chosen by Fly Launch, or select a different region to deploy to.
+
+* Services: The port for services (the value of `internal_port` in the `fly.toml` file) depends on the `EXPOSE` instruction in your Dockerfile. The default port when the `fly launch` command doesn't find ports set in a Dockerfile is `8080`.
```
andie787, line 41: we should be specific about what is configured for this example...should they set it to 80 here? Or don't mention services here and leave the instruction for fly.toml below?

languages-and-frameworks/wordpress.html.markerb
```diff
@@ -0,0 +1,160 @@
...
+
+* You can also set a default Machine size and memory, and add and configure Postgres or Redis.
```
andie787, line 43: Is this sentence needed here, they won't be adding pg or changing the memory?

languages-and-frameworks/wordpress.html.markerb
```diff
@@ -0,0 +1,160 @@
...
+
+-------
+ ✔ [1/2] Machine 918543b477de83 [app] update succeeded
+ ✔ [2/2] Machine e28697ce6d3986 [app] update succeeded
+-------
+
+Visit your newly deployed app at https://my-app-name.fly.dev/
+```
+
+### Open your app
```
andie787, line 144: Do you think a lot of wordpress users would be using a custom domain? We can [link to that](https://fly.io/docs/networking/custom-domain/) or even include a quick version here?

andie787 (commented):
This is a really good addition to docs!

Did a quick review and left some suggestions.

andie787 (commented):
Nice! Merge away whenever you're ready. 🚀

PR #1669, opened 2024-07-12
tigris documentation

senyosimpson:
### Summary of changes

Documentation for serving files with Tigris, an alternative method for serving statics


reference/configuration.html.markerb
```diff
@@ -958,13 +958,32 @@ Each `[[statics]]` block maps a URL prefix to a path inside your container. You
...
+```
+
 ### Caveats
 
 This feature should not be compared directly with a CDN, for the following reasons:
 
 * This feature does not exempt you from having to run a web service in your Machine.
 
-* Statics will not find `index.html` at the root. The full path must be requested.
+* Statics will not find `index.html` at the root, except when serving statics from Tigris directly. The full path must be requested.
```
andie787, line 986: mention `index_document` here again?

andie787 (commented):
This is looking good to me! Do you think we can update the main Tigris doc too? https://fly.io/docs/reference/tigris/

PR #1671, opened 2024-07-12
Regions: add edges

andie787:
### Summary of changes

One option for adding edge-only regions to the regions table. The other option is to have them listed separately as a section below the region table.

### Preview

https://aa-docs-2.fly.dev/docs/reference/regions/

### DIFF
--- a/reference/regions.html.markerb
+++ b/reference/regions.html.markerb
@@ -28,7 +28,6 @@ You can host your apps in any of the following regions.
 | cdg       | Paris, France                 | ✓           |                                                       |
 | den       | Denver, Colorado (US)         |             |                                                       |
 | dfw       | Dallas, Texas (US)            | ✓           |                                                       |
-| dxb       | Dubai, United Arab Emirates   |             |                                                       |
 | ewr       | Secaucus, NJ (US)             |             |                                                       |
 | eze       | Ezeiza, Argentina             |             |                                                       |
 | fra       | Frankfurt, Germany            | ✓           | ✓                                                     |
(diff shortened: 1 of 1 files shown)


gargakshit (approved):

reference/regions.html.markerb
```diff
@@ -14,51 +14,67 @@ Run <code>fly platform regions</code> to get a list of regions.
...
+| yyz       | Toronto, Canada               | ✓           |                                                       |
+
+- **Gateway regions:** "Gateway" regions also have WireGuard gateways, through which you connect to your organization's private network.
+- **Launch Plan or higher regions:** For some higher-demand regions, we restrict scaling up Machines to organizations with the [Launch, Scale, or Enterprise plans](https://fly.io/plans).
+
+## Edge-only regions
+
+Edge servers receive inbound traffic from the Internet destined for apps on the Fly.io platform and private traffic between Machines in different regions. [Fly Proxy](/docs/reference/fly-proxy/) routes traffic from edges to regional worker servers.
+
+We have a few regions that don't have workers and aren't available for app deployment:
```
asib, line 65:
```suggestion
Edge servers receive inbound traffic from the Internet destined for apps on the Fly.io platform and private traffic between Machines in different regions. [Fly Proxy](/docs/reference/fly-proxy/) routes traffic from edges to regional worker servers. Edge-only regions don't have any workers and therefore aren't available for app deployment.
```

Small suggestion to move that last sentence into the paragraph above - it felt a bit like a non sequitur as a separate sentence.

asib (approved):
One small suggestion, but otherwise looks great :)

PR #1680, opened 2024-07-14
Expand Elixir clustering nodes page

brain-geek:
### Summary of changes
- Clarify that RELEASE_COOKIE can be not only in `.toml` file, but also in secrets
- Improve transition to `Troubleshooting` section

### Preview
Screenshot 2024-07-14 at 17 57 04

- https://fly.io/docs/elixir/the-basics/clustering/

### Notes
The original version was confusing in two aspects:
 - It mentioned that cookie should be set in `.toml` file without mentioning alternatives.
 - It implied going to the next step (`...we are ready for the next step!`) while next step was `Troubleshooting` which is optional.
 - Code generating cookie value was after the explanation how to set it.


andie787:
@brain-geek don't worry too much about the Vale linter. Sometimes it complains when it doesn't recognize a word. :)

PR #1682, opened 2024-07-15
Add doc to move app to a new org

andie787:
### Summary of changes

Move an app to a new org as suggested and started by @Lucais11

### DIFF
--- a/apps/move-app-org.html.markerb
+++ b/apps/move-app-org.html.markerb
@@ -1,0 +1,38 @@
+---
+title: Move an app between organizations
+layout: docs
+nav: apps
+---
+
+You might want to move an app to a new org to hand off your completed work to a client, to take advantage of a new pricing model, or to move a production app out of your personal organization.
+
+You can move an entire app with its resources to another organization with the `fly apps move` command. You'll need to be a member of both organizations to move an app.
+
+To move an app from one organization to another:
+
+```cmd
+fly apps move <app name> --org <target organization name>
+```
+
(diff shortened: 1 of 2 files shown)


Lucais11 (approved):
When moving over LiteFS, if their volume data & secrets are all being copied and they're using $FLY_APP_NAME in their Consul key, it should "just work". Do we want to make a note in the doc about that to save people from possibly running into that issue?

Other than that, this looks great!!!

andie787:
> When moving over LiteFS, if their volume data & secrets are all being copied and they're using $FLY_APP_NAME in their Consul key, it should "just work". Do we want to make a note in the doc about that to save people from possibly running into that issue?
> 
> Other than that, this looks great!!!

@Lucais11 The process does "copy" the volume and secrets, right? So is the key to this statement just the $FLY_APP_NAME variable in the Consul key?

Lucais11:
> > When moving over LiteFS, if their volume data & secrets are all being copied and they're using $FLY_APP_NAME in their Consul key, it should "just work". Do we want to make a note in the doc about that to save people from possibly running into that issue?
> > Other than that, this looks great!!!
> 
> @Lucais11 The process does "copy" the volume and secrets, right? So is the key to this statement just the $FLY_APP_NAME variable in the Consul key?

AFAIK yes it does copy the volume and secrets and so we would just want to add the statement about the $FLY_APP_NAME variable in the Consul key.

Lucais11 (approved):

PR #1693, opened 2024-07-17
Meteor sticky sessions with `fly-replay`

kylemclaren:

### DIFF
--- a/js/frameworks/meteor.html.markerb
+++ b/js/frameworks/meteor.html.markerb
@@ -198,17 +198,46 @@ Opening https://hello-meteor.fly.dev/
 
 Your browser will be sent to the displayed URL.
 
-## _A word on Session Affinity (AKA 'Sticky Sessions')_
+## _Session Affinity (aka 'Sticky Sessions')_
 
 Sticky sessions are a feature of a load balancer that ensures that all requests from a client are routed to the same server. This can be important for Meteor applications depending on the client/server architecture of your app, as it ensures that the client is always connected to the same server, and that any changes made to the server are reflected in the client.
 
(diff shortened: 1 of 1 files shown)


rubys (approved):

js/frameworks/meteor.html.markerb
```diff
@@ -198,18 +198,47 @@ Opening https://hello-meteor.fly.dev/
 
 Your browser will be sent to the displayed URL.
 
-## _A word on Session Affinity (AKA 'Sticky Sessions')_
+## _Session Affinity (aka 'Sticky Sessions')_
 
 Sticky sessions are a feature of a load balancer that ensures that all requests from a client are routed to the same server. This can be important for Meteor applications depending on the client/server architecture of your app, as it ensures that the client is always connected to the same server, and that any changes made to the server are reflected in the client.
 
-To achieve this on Fly.io, you can run your app with a maximum of one Machine per region. The [Fly Proxy](docs/reference/fly-proxy/) will take care of load balancing clients to the Machine closest to them, and will ensure that all requests from a client are routed to the same Machine.
```
rubys, line 205: I'm uncomfortable with the word *ensure*.  The internet is a weird place.  I'm in Raleigh NC, and normally my entrypoint is at iad.  But if frequently is atl.  And periodically there are times when it is dfw.  Limiting the number of servers in a region, and spacing the regions far apart make it more *likely* subsequent requests will arrive at the same server, but doesn't *guarantee* that to be the case.

kylemclaren: Good call, cleaned that up to clarify 'best effort' behavior

PR #1701, opened 2024-07-19
Machine reservations docs

asib:
### Summary of changes

Docs for [machine reservations](https://community.fly.io/t/reservation-blocks-40-discount-on-machines-when-youre-ready-to-commit/20858).

### Preview

image

https://community.fly.io/t/reservation-blocks-40-discount-on-machines-when-youre-ready-to-commit/20858

### Notes

N/A

### DIFF
--- a/about/pricing.html.markerb
+++ b/about/pricing.html.markerb
@@ -154,9 +154,10 @@ regionSelect.addEventListener("change", function () {
 
 For stopped Machines we charge only for the root file system (rootfs) needed for each Machine. Each 1GB of rootfs for a Machine stopped for 30 days is $0.15. The amount of rootfs needed is defined by your OCI image generated on your app plus a few [containerd](https://containerd.io/+external) tweaks on the underlying file system.
 
-### Machine Reservation Blocks
+### Machine reservation blocks
 
-We have a block-based commit scheme. Reserve a block of compute time, either for `performance` Machines or `shared` Machines, for a 40% discount. Reservations apply to any number of Machines in that class, in a specific region.
+You'll get a 40% discount when you reserve a block of compute time, either for `performance` Machines or `shared` Machines, in a specific region.
+Reservations apply to any number of Machines in that class.
 
 The available reservation sizes are:
 
--- a/about/pricing.html.markerb
+++ b/about/pricing.html.markerb
@@ -173,7 +173,9 @@ The available reservation sizes are:
 * $360/yr for $50/mo of usage
 * $3,600/yr for $500/mo of usage
 
-There's no limit on the number or combinations of blocks that can be purchased. Reservations are backdated to the first day of the current month.
+You pay the "per year" amount upfront, and each month receive a credit worth the "per month" amount. The credit does not rollover; it's only valid for the month in which it's granted.
+
+There's no limit on the number or combinations of blocks that can be purchased. Reservations are backdated to the first day of the month in which they're purchased.
 
 Email reservations@fly.io with a list of the reservations you’d like to purchase along with which regions, and we’ll set them up for you.
 
--- a/about/pricing.html.markerb
+++ b/about/pricing.html.markerb
@@ -163,13 +163,13 @@ The available reservation sizes are:
 
 **Performance Machines**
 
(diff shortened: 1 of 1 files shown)


about/pricing.html.markerb
```diff
@@ -154,6 +154,28 @@ regionSelect.addEventListener("change", function () {
...
+
+The available reservation sizes are:
+
+**Performance Machines**
+
+* $144/yr for $20/mo of usage (i.e. $12/mo amortised with a 12 month commit)
+* $1,440/yr for $200/mo of usage
+* $14,400/yr for $2,000/mo of usage
+
+**Shared Machines:**
```
andie787, line 169:
```suggestion
**Shared Machines**
```

about/pricing.html.markerb
```diff
@@ -154,6 +154,28 @@ regionSelect.addEventListener("change", function () {
...
 
+### Machine Reservation Blocks
+
+We have a block-based commit scheme. Reserve a block of compute time, either for `performance` Machines or `shared` Machines, for a 40% discount. Reservations apply to any number of Machines in that class, in a specific region.
+
+The available reservation sizes are:
+
+**Performance Machines**
+
+* $144/yr for $20/mo of usage (i.e. $12/mo amortised with a 12 month commit)
```
andie787, line 165: I think it might be better to explain this: `(i.e. $12/mo amortised with a 12 month commit)` in the description above rather than inline on only one price. Something like:

> We amortize the yearly cost over the year and you make a 12 month commitment. For example, if you buy a reservation block of $360/yr for $50/mo of usage, you'll pay $30/mo for 12 months.

about/pricing.html.markerb
```diff
@@ -154,6 +154,28 @@ regionSelect.addEventListener("change", function () {
 
 For stopped Machines we charge only for the root file system (rootfs) needed for each Machine. Each 1GB of rootfs for a Machine stopped for 30 days is $0.15. The amount of rootfs needed is defined by your OCI image generated on your app plus a few [containerd](https://containerd.io/+external) tweaks on the underlying file system.
 
+### Machine Reservation Blocks
+
+We have a block-based commit scheme. Reserve a block of compute time, either for `performance` Machines or `shared` Machines, for a 40% discount. Reservations apply to any number of Machines in that class, in a specific region.
```
andie787, line 159: "scheme" has negative connotations (maybe esp in the US? I don't know)

```suggestion
You can purchase resource block-based commitments. You'll get a 40% discount when you reserve a block of compute time, either for `performance` Machines or `shared` Machines in a specific region.
```
or something like that? Maybe you'll need to add back in the bit about "no limit to the number of Machines"? If I understood that correctly?

andie787: or do we want to say "purchase reservation" as below rather than "commitments"? In any case we should choose one of "reservation", "block", or "commitment" as the thing being purchased.

andie787 (commented):
left some comments re: figuring out naming and so on

about/pricing.html.markerb
asib: Ah we don't actually do that, you pay the $360 upfront. But I get the point, we should put this in proper prose.

andie787:
@asib do you want to take a crack re-wording/clarifying? When you're happy with it I'm fine with you just merging!

nbw:
`Machine Reservation Blocks` is already a long title, but do we want to include a mention of it being region specific? Or could we mention the "region-specfic"-ness in the first sentence rather than the end of the first paragraph?

> You can purchase resource block-based commitments on a per-region basis.

asib:
> `Machine Reservation Blocks` is already a long title, but do we want to include a mention of it being region specific? Or could we mention the "region-specfic"-ness in the first sentence rather than the end of the first paragraph?
> 
> > You can purchase resource block-based commitments on a per-region basis.
> 
> imho, the value proposition is ultimately discounted machines _for a single region_ so that should be obvious before proceeding.

Yeh great note, I'll go with your suggestion.

about/pricing.html.markerb
asib: Rejigged. Essentially got rid of the first sentence, since it felt like it was repetitive given content of second sentence.

PR #1702, opened 2024-07-20
Updates for Machine suspension

matttpt:
### Summary of changes

Update documentation to account for the new Machine suspension feature, including proxy autosuspend:

* Reference the new `suspending` and `suspended` Machine states
* Document the `suspend` Machines API endpoint
* Explain how the `start` and `stop` Machines API endpoints affect `suspended` Machines
* Document new autostop settings in the Machines API and in `fly.toml`

* https://community.fly.io/t/new-feature-in-preview-suspend-resume-for-machines/20672

### Notes

Still a work in progress right now!

### DIFF
--- a/machines/api/machines-resource.html.markerb
+++ b/machines/api/machines-resource.html.markerb
@@ -313,7 +313,7 @@ Create a Machine with services defined on app. Learn more about [services and ne
     <% api_info.add_info(
       name: 'state',
       type: 'string enum',
-      description: 'The Machine state to wait for. Values are: `started`, `stopped`, or `destroyed`. Default is `started`.'
+      description: 'The Machine state to wait for. Values are: `started`, `stopped`, `suspended`, or `destroyed`. Default is `started`.'
     ) %>
     <%= render(api_info) %>
     <% api_info = ApiInfoComponent.new(
--- a/machines/api/machines-resource.html.markerb
+++ b/machines/api/machines-resource.html.markerb
@@ -479,7 +479,7 @@ This is, in particular, how you would update the running image of a Machine (whe
 
 ## Stop a Machine
 
-Stopping a started Machine will shut down the Machine, but not destroy it. The Machine may be started again with `machines/<machine_id>/start`.
+Stopping a started Machine will shut down the Machine, but not destroy it. The Machine can be started again with `machines/<machine_id>/start`.
 
 Stopping a suspended Machine will invalidate its snapshot, forcing it to perform a cold boot the next time it is started.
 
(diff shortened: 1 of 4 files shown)


machines/api/machines-resource.html.markerb
```diff
@@ -479,7 +479,9 @@ This is, in particular, how you would update the running image of a Machine (whe
 
 ## Stop a Machine
 
-Stopping a Machine will shut down the Machine, but not destroy it. The Machine may be started again with `machines/<machine_id>/start`.
+Stopping a started Machine will shut down the Machine, but not destroy it. The Machine may be started again with `machines/<machine_id>/start`.
```
andie787, line 482: this wasn't you, but I think we should change "may" to "can"
```suggestion
Stopping a started Machine will shut down the Machine, but not destroy it. The Machine can be started again with `machines/<machine_id>/start`.
```

machines/api/machines-resource.html.markerb
```diff
@@ -528,9 +530,60 @@ Stopping a Machine will shut down the Machine, but not destroy it. The Machine m
...
+</div>
+
 ## Start a Machine
 
-Start a stopped Machine. Machines that are restarted are completely reset to their original state so that they start clean on the next run.
+Start a Machine.
+
+Stopped Machines that are restarted are completely reset to their original state so that they start clean on the next run.
+
+Suspended Machines that are started attempt to resume from the snapshot taken when they were suspended. If this is not possible, then they will perform a normal cold boot.
```
andie787, line 586: Suggestion (if you can make it less awkward then go for it :))
```suggestion
Suspended Machines that are started attempt to resume from the snapshot taken when they were suspended. If this is not possible, then they will perform a normal cold boot, as though starting from the stopped state.
```

andie787 (commented):
looking good so far! Left a few comments. When you're happy, feel free to merge!

andie787:
when you're happy with the updates, then you can merge and we can always iterate on additions/improvements!

machines/api/machines-resource.html.markerb
matttpt: @andie787 LGTM; I'm not sure if I can find a better way to phrase it!

One thing I realized is that (currently) the Machine's root file system may not be reset when this happens, in which case it isn't "completely reset to [its] original state," so not *exactly* as though starting from `stopped`. Does that deserve a mention here, or will it muddle this too much?

(Long-term, it may be a good idea to make this behavior more consistent …)

andie787: If we know exactly when it does or doesn't, then it might be worth mentioning? We try to give all the needed context in-place in the API docs.

matttpt: After thinking through it more: right now I believe that the root file system will *never* be reset when this happens. Added this as a caveat.

PR #1714, opened 2024-07-25
Request headers: update to correct inaccurate info about client IP

### LINKED ISSUE #1554: Issue with the "Request headers" doc
I found an issue with this document.

Title: Request headers
Location: https://fly.io/docs/networking/request-headers/
Source: https://github.com/superfly/docs/blob/main/networking/request-headers.html.md

### Describe the issue

This statement, "will also appear at the start of", from below snipppet, is misleading :

> #### `Fly-Client-IP`
> **Client IP Address:** The IP address the Fly Proxy accepted a connection from. This will be the client making the initial request and as such, will also appear at the start of the `X-Forwarded-For` addresses.

While it is true that for most users, most of the time, the left-most IP will be the user's IP, this is not a truthful statement, and may be dangerous if developers read this and write code based on it.

It is more accurate to say that the right-most IP preceding any load balancers/reverse proxies is the users IP.

For example, consider situations:

#### If the user spoofs their IP

```
X-Forwarded-For: {Spoofed IP}, {Real Remote Address}, {Fly.io Reverse Proxy}, {NGINX Hosted on Fly.io}
                               # ^ this is hopefully appended by the Fly.io Reverse Proxy,
                               # before it's own IP is appended
```

#### If the user is behind a forward proxy

```
X-Forwarded-For: {User IP}, {Forward Proxy eg. School Firewall}, {Fly.io Reverse Proxy}, {NGINX Hosted on Fly.io}
                            # ^ this should be the Fly-Client-IP, as there is no way to 
                            # know that the User IP was/wasn't spoofed
```

Hence, "will also appear at the start of the `X-Forwarded-For` addresses" is not an accurate statement

### Addtional info

I am assuming this is just an issue with the docs - I have not tested this code. Would be good if one of the Fly devs could verify the implementation of the Fly Proxy, since I don't believe it's open source

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#proxy1_proxy2

andie787:
### Summary of changes

- update the description of `Fly-Client-IP` to clarify what happens when you use a proxy, and to link to to the `X-Forwarded-For` description
- update the description of `X-Forwarded-For` for clarify why you want to go read the MDN docs

- fixes #1554 
- https://community.fly.io/t/minor-doc-correction/20944

### DIFF
--- a/networking/request-headers.html.md
+++ b/networking/request-headers.html.md
@@ -9,7 +9,7 @@ Request headers carry information that is specific to the incoming request and i
 ## Request Headers
 
 ### `Fly-Client-IP`
-**Client IP Address**: The IP address the Fly Proxy accepted a connection from. You can use `Fly-Client-IP` to get the IP address of the client making the initial request if you aren't using any other proxy service for your app in front the Fly.io platform. When you use a proxy, such as Cloudflare, requests will appear to come from the proxy service's IP addresses, rather than the client making the request. In that case, you'll need to parse the IP addresses in the [`X-Forwarded-For` header](#x-forwarded-for) to find the client IP address.
+**Client IP Address**: The IP address of the client from the perspective of Fly Proxy. If you are using another reverse proxy service in front of the Fly.io platform, this header will be set to proxy service's IP addresses, rather than the client making the request. In that case, you'll need to parse the IP addresses in the [`X-Forwarded-For` header](#x-forwarded-for) to find the client IP address.
 
 ### `Fly-Forwarded-Port`
 **Original connection port**: This header is always set by the Fly Proxy and denotes the actual port that the client connected to the Fly edge node which is then forwarded to the application instance.
@@ -20,9 +20,9 @@ Request headers carry information that is specific to the incoming request and i
 Not to be confused with the [environment variable](/docs/machines/runtime-environment/#fly_region) `FLY_REGION`, which is where the application is running.
 
 ### `X-Forwarded-For`
-**Client and Proxy List**: A comma separated list of IP addresses including the address of the client that originated the request and the addresses of the proxy servers the request passed through. For example, "77.97.0.98, 77.83.142.33" contains the client and the one proxy it passed through.
+**Client and Proxy List**: A comma separated list of IP addresses including the address of the client that originated the request and the addresses of the proxy servers the request passed through. For example, "77.97.0.98, 77.83.142.33" contains the client and the one proxy it passed through. On the Fly.io platform, the last address (rightmost) in this list will be a shared or dedicated IP address assigned to your app.
 
-Follow the directives in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For+external) for `X-Forwarded-For` to learn how to avoid selecting a spoofed IP address or a proxy service IP address as the client IP.
+This header must be treated with caution to avoid spoofing attempts. Follow the directives in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For+external) for `X-Forwarded-For` to learn how parse the header safely. If you aren't using another reverse proxy in front of the Fly.io platform, the `Fly-Client-IP` might be a better choice to get client IP addresses.
 
 ### `X-Forwarded-Proto`
 **Original client protocol**: The protocol which the client used to make the request. Either `http` or `https`.
--- a/networking/request-headers.html.md
+++ b/networking/request-headers.html.md
@@ -9,30 +9,39 @@ Request headers carry information that is specific to the incoming request and i
 ## Request Headers
 
 ### `Fly-Client-IP`
+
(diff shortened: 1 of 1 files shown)


networking/request-headers.html.md
```diff
@@ -9,7 +9,7 @@ Request headers carry information that is specific to the incoming request and i
 ## Request Headers
 
 ### `Fly-Client-IP`
-**Client IP Address**: The IP address the Fly Proxy accepted a connection from. This will be the client making the initial request and as such, will also appear at the start of the `X-Forwarded-For` addresses. 
+**Client IP Address**: The IP address the Fly Proxy accepted a connection from. You can use `Fly-Client-IP` to get the IP address of the client making the initial request if you aren't using any other proxy service for your app in front the Fly.io platform. When you use a proxy, such as Cloudflare, requests will appear to come from the proxy service's IP addresses, rather than the client making the request. In that case, you'll need to parse the IP addresses in the [`X-Forwarded-For` header](#x-forwarded-for) to find the client IP address.
```
andie787, line 12: I'm not sure this is accurate... is this also the case if the user accessing the app is spoofing their own IP?

PeterCxy: @andie787 A parser for `X-Forwarded-For` as prescribed by the MDN doc shouldn't be fooled by a user spoofing their own IP based on that header, because it should stop at any point in that header where it sees an untrusted IP.

I'm wondering whether something like this would work better:

> **Client IP Address**: The IP address of the client _from the perspective of Fly Proxy_. If you are using another reverse proxy service in front of the Fly.io platform, this header will be set to the proxy service's IP address, rather than the client making the request. In that case, you'll need to parse the IP addresses in the [`X-Forwarded-For` header](#x-forwarded-for) to find the client IP address.

networking/request-headers.html.md
```diff
@@ -20,9 +20,9 @@ Request headers carry information that is specific to the incoming request and i
 Not to be confused with the [environment variable](/docs/machines/runtime-environment/#fly_region) `FLY_REGION`, which is where the application is running.
 
 ### `X-Forwarded-For`
-**Client and Proxy List**: This is a comma separated list comprising of the client that originated the request and the proxy servers the request passed through. For example, "77.97.0.98, 77.83.142.33" contains the client and the one proxy it passed through.
+**Client and Proxy List**: A comma separated list of IP addresses including the address of the client that originated the request and the addresses of the proxy servers the request passed through. For example, "77.97.0.98, 77.83.142.33" contains the client and the one proxy it passed through.
```
PeterCxy, line 23: What about also adding:

> On the Fly.io platform, the last address in this list will be a shared or dedicated IP address assigned to your app.

networking/request-headers.html.md
```diff
@@ -20,9 +20,9 @@ Request headers carry information that is specific to the incoming request and i
...
 
-MDN has [full documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) for this header.
+Follow the directives in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For+external) for `X-Forwarded-For` to learn how to avoid selecting a spoofed IP address or a proxy service IP address as the client IP.
```
PeterCxy, line 25: I think we could make it clearer that blindly trusting this header is dangerous, and direct the user to use `Fly-Client-IP` if they don't want to deal with all this complexity, maybe:

> This header __must be treated with caution__ to avoid spoofing attempts. Follow the directives in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For+external) to learn how to parse the header safely. If you aren't using another reverse proxy in front of the Fly.io platform, the `Fly-Client-IP` may be an easier choice to get access to client IP addresses.

andie787:
@PeterCxy thanks for your review; took all your suggestions!

networking/request-headers.html.md
```diff
@@ -20,9 +20,9 @@ Request headers carry information that is specific to the incoming request and i
...
+**Client and Proxy List**: A comma separated list of IP addresses including the address of the client that originated the request and the addresses of the proxy servers the request passed through. For example, "77.97.0.98, 77.83.142.33" contains the client and the one proxy it passed through. On the Fly.io platform, the last address (rightmost) in this list will be a shared or dedicated IP address assigned to your app.
 
-MDN has [full documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) for this header.
+This header must be treated with caution to avoid spoofing attempts. Follow the directives in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For+external) for `X-Forwarded-For` to learn how parse the header safely. If you aren't using another reverse proxy in front of the Fly.io platform, the `Fly-Client-IP` might be a better choice to get client IP addresses.
```
PeterCxy, line 25: Just realized I missed a "header" here after `Fly-Client-IP` :) Maybe we should link that to the section for it too?

andie787:
@PeterCxy also a question: "On the Fly.io platform, the last address in this list will be a shared or dedicated IP address assigned to your app." Is this the leftmost? I had assumed rightmost at first based on "last"

PeterCxy:
@andie787 It's the rightmost. I believe all proxies are supposed to add their own addresses to the very end of that list.

PR #1717, opened 2024-07-25
do more section for python

fliepeltje:
### Summary of changes

- abstracts some core paths into partials
- adds a 'do more' section for python that uses these partials

### Preview

https://green-brook-5576.fly.dev/docs/python/do-more/

### Notes

Each of the sections in pyton/do-more 'borrows' content from the parent, for example you will note the start of [the postgres docs](https://green-brook-5576.fly.dev/docs/postgres/getting-started/create-pg-cluster/) is the same as the [python postgres docs](https://green-brook-5576.fly.dev/docs/python/do-more/add-postgres/) - they share the same partial for the language-independent part. That way if anything changes about `fly pg create` the downstream docs remain up to date.


gpus/partials/_ollama-create.erb
```diff
@@ -0,0 +1,49 @@
...
+
+[[vm]]
+  size = 'a100-80gb'
+```
+
+Modify the app name and region to suit your preferences and save the file as `fly.ollama.toml`.
+Then you can launch the app:
+
+```cmd
+fly launch -c fly.ollama.toml --flycast
```
andie787, line 31: Whoa! A flyctl option for flycast. Nice.

gpus/partials/_ollama-create.erb
```diff
@@ -0,0 +1,49 @@
...
+```
+
+There's a couple of things to note here:
```
andie787, line 34:
```suggestion
There are a couple of things to note here:
```

gpus/partials/_ollama-create.erb
```diff
@@ -0,0 +1,49 @@
...
+Finally, this _only_ starts the `ollama` server; at this point you cannot interact with any models yet.
+To do so, you will have to pull in a model with this one easy, short, intuitive command:
+
+```cmd
+fly m run -e OLLAMA_HOST=http://<your-app>.flycast --shell --command "ollama pull llama3.1" ollama/ollama
+```
+
+This command will pull in the `llama3.1` model, change the model name to suit your needs.
+At this point this model is now available to the internal network of the organization it is deployed in.
+You can access it using flycast from this URL: `http://<your-app>.flycast`.
```
andie787, line 49:
```suggestion
You can access it using Flycast from this URL: `http://<your-app>.flycast`.
```

gpus/partials/_ollama-create.erb
```diff
@@ -0,0 +1,49 @@
...
+```cmd
+fly launch -c fly.ollama.toml --flycast
+```
+
+There's a couple of things to note here:
+- The `ollama` image is a GPU image, so you need to specify a GPU size in the `[[vm]]` section.
+- Not all regions have GPUs available, refer to [region GPU availability](/docs/gpus/#regions-with-gpus) for more info
+- The `[[mounts]]` section is used to mount a volume to store the models. This is a good practice to keep the models separate from the app code.
+- The volume size in this config file is 10gb, that's enough for small models. Change this value if you need more.
+
```
andie787, line 39: Add a bullet for the `--flycast` option creating a private IPv6 address?

gpus/partials/_ollama-create.erb
```diff
@@ -0,0 +1,49 @@
+Running `ollama` on a GPU is pretty easy. The fastest way to get started is to adapt the following `fly.toml` file:
```
andie787, line 1: **For all the partials:** We want to move all the partials under the main `partials` directory, with sub-directories there, so that all the partials are easy to find. For this PR, could you move any new ones you're creating? (No worries about existing partials, I'll fix those later!)
partials/tigris/
partials/postgres/
partials/gpus/

reference/partials/_tigris-storage-create.erb
```diff
@@ -0,0 +1,36 @@
+## Create and manage a Tigris storage bucket
+
+Creating and managing storage buckets happens exclusively via the [Fly CLI](/docs/flyctl/install/). Install it, then [sign up for a Fly account](/docs/getting-started/sign-up-sign-in/).
```
andie787, line 3: I don't think this sentence fits here in the context of Do More. 

You could fix this by excluding the title and the first sentence from the partial and then adding a custom intro to the object storage python page.

andie787:
Merge whenever you're ready to give this a go!

PR #1718, opened 2024-07-25
Add section on different ways to authenticate to the Prometheus API

Roadmaster:
### Summary of changes
* Explain how to create a non-bearer macaroon token with restrictions (per org, read-only)
* Explain how to pass a macaroon token in the Authorization header

This process is known internally and we share it with customers who require it but a customer pointed out it's not documented which makes it difficult to discover.

### DIFF
--- a/monitoring/metrics.html.md
+++ b/monitoring/metrics.html.md
@@ -412,7 +412,7 @@ Create an org-restricted token:
 fly token create org -o THE_ORGANIZATION
 ```
 
-A read-only token is created like this:
+Create a read-only org-restricted token:
 
 ```
 fly token create readonly
--- a/monitoring/metrics.html.md
+++ b/monitoring/metrics.html.md
@@ -388,3 +388,48 @@ processes = ["proxy"]
 Instrument your app and expose your metrics on `0.0.0.0`.
 
 There are many supported [client libraries](https://prometheus.io/docs/instrumenting/clientlibs/) as well as off-the-shelf [exporters](https://prometheus.io/docs/instrumenting/exporters/) able to return Prometheus-formatted metrics.
+
+## Authentication
+
+Authenticating to the Prometheus API can be achieved a few different ways, depending on the level of access you want your token to have.
+
+### Fly Access Token
+
+As in the earlier example, a full access token can be generated with `flyctl auth token` and then passed as a bearer token in the `Authorization` header. The header looks like:
+
+```
+Authorization: Bearer THE_TOKEN
+```
+
+### Fly org-restricted or read-only token
+
+This kind of token or "[macaroon](https://fly.io/blog/macaroons-escalated-quickly/)" can be scoped to a single organization or configured to only allow read operations, which can be safer than using a full-blown read-write token that grants access to all organizations under your account.
+
+### Generating
+An org-restricted token can be generated like so:
+
+```
+fly token create org -o THE_ORGANIZATION
+```
+
+A read-only token is created like this:
+
+```
+fly token create readonly
+```
+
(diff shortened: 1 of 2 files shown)


monitoring/metrics.html.md
```diff
@@ -112,7 +112,7 @@ the user icon in the lower-left of the screen.
 
 ### External or self-hosted Grafana
 
-You can also configure your Prometheus endpoint with an existing Grafana installation, or [host one on Fly.io](https://github.com/fly-apps/grafana). Either way, you set it up thusly:
+You can also configure your Prometheus endpoint with an existing Grafana installation, or [host one on Fly.io](https://github.com/fly-apps/grafana). Either way, you set it up thus:
```
andie787, line 115:
```suggestion
You can also configure your Prometheus endpoint with an existing Grafana installation, or [host one on Fly.io](https://github.com/fly-apps/grafana). Either way, you can set it up like this:
```

monitoring/metrics.html.md
```diff
@@ -388,3 +388,48 @@ processes = ["proxy"]
...
+
+```
+Authorization: Bearer THE_TOKEN
+```
+
+### Fly org-restricted or read-only token
+
+This kind of token or "[macaroon](https://fly.io/blog/macaroons-escalated-quickly/)" can be scoped to a single organization or configured to only allow read operations, which can be safer than using a full-blown read-write token that grants access to all organizations under your account.
+
+### Generating
```
andie787, line 408:
```suggestion
### Generating tokens
```

monitoring/metrics.html.md
```diff
@@ -388,3 +388,48 @@ processes = ["proxy"]
...
+An org-restricted token can be generated like so:
```
andie787, line 409:
```suggestion
Create an org-restricted token:
```

monitoring/metrics.html.md
```diff
@@ -388,3 +388,48 @@ processes = ["proxy"]
...
+
+```
+fly token create org -o THE_ORGANIZATION
+```
+
+A read-only token is created like this:
```
andie787, line 415:
```suggestion
Create a read-only org-restricted token:
```

monitoring/metrics.html.md
```diff
@@ -388,3 +388,48 @@ processes = ["proxy"]
...
+
+Note that the read-only token still has access to all organizations in your account (although it cannot modify anything, only read data). These two can be combined to create a read-only token that is restricted to a single organization:
+
+```
+fly token create readonly --from-existing -t "$(fly token create org -o THE_ORGANIZATION)"
+```
+
+These tokens look like this once generated: `FlyV1 fm2_lJPECAAAAAAAAC7txBAzYI6PRWhHLT...(a lot of base64-encoded text)`.
+
+#### Using
```
andie787, line 429:
```suggestion
#### Using tokens
```

monitoring/metrics.html.md
```diff
@@ -388,3 +388,48 @@ processes = ["proxy"]
...
+fly token create org -o THE_ORGANIZATION
+```
+
+A read-only token is created like this:
+
+```
+fly token create readonly
+```
+
+Note that the read-only token still has access to all organizations in your account (although it cannot modify anything, only read data). These two can be combined to create a read-only token that is restricted to a single organization:
```
andie787, line 421: The read-only token you create with `fly token create readonly` is already an org scoped token, so you don't need to do this.

andie787 (commented):
Thanks for adding this info! Left some feedback to consider.

monitoring/metrics.html.md
Roadmaster: Thanks, I'll remove this part entirely :)

PR #1723, opened 2024-07-26
Fix example YAML for GitHub Review Apps by bumping `superfly/fly-pr-review-apps` version to latest

skalnik:
### Summary of changes

I recently tried to follow this and ran into a few stumbling blocks. 

1. ~~The `steps` property is missing, so the YAML wasn't valid~~ Looks like this was fixed in another PR
2. `superfly/fly-pr-review-apps` was updated, the current version used (`1.2.0`) uses a flag that no longer exists in the flyctl, and thus this errors

https://github.com/superfly/fly-pr-review-apps/pull/56

### Notes

There's also a link to a gist on Line 21, which should be updated to the latest version of `superfly/fly-pr-review-apps` as well

### DIFF
--- a/blueprints/review-apps-guide.html.md
+++ b/blueprints/review-apps-guide.html.md
@@ -82,13 +82,13 @@ jobs:
       name: review
       # The script in the `deploy` sets the URL output for each review app.
       url: ${{ steps.deploy.outputs.url }}
-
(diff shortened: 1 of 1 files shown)


andie787:
thanks @skalnik !

PR #1737, opened 2024-08-05
Fix typo on rails getting-started

galori:
It said "respond with "Y" or press enter" to the question of "do you want to make change) but the default is "N" and it says changes aren't needed, so clearly the intent was "respond with "Y""


andie787:
You're right! Thanks

Issue #1738, opened 2024-08-05
Issue with the "Run a FastAPI app" doc + "fastapi-cli" missed


icehongssii:
I found an issue with this document.

Title: Run a FastAPI app
Location: https://fly.io/docs/python/frameworks/fastapi/
Source: https://github.com/superfly/docs/blob/main/python/frameworks/fastapi.html.markerb

### Describe the issue

The official doc said `We can then serve the development version of the app using the fastapi cli tool - fastapi run`.
But the doc has never mentioned installing fastapi-cli.
Thus,  `poetry add fastapi`has to include `fastapi-cli`

```
poetry add fastapi fastapi-cli
```

### Addtional info

![image](https://github.com/user-attachments/assets/fdbb2039-0fb3-408e-b45c-64af1443ec7c)

andie787:
I think the other option is to run `poetry add "fastapi[standard]"`. 
@fliepeltje what is your preference?

fliepeltje:
This has changed in the most recent version of FastAPI; they've changed to not include the CLI by default, so yeah we should update it. Thanks for catching it!

PR #1740, opened 2024-08-07
Add fastapi-cli while instaling fastapi

viva-douner:
### Summary of changes

The official doc said We can then serve the development version of the app using the fastapi cli tool - `fastapi run`.
But the doc has never mentioned installing fastapi-cli.
Thus, `poetry add fastapi` has to include fastapi-cli

`poetry add fastapi fastapi-cli`

### Preview

image

#1738


kcmartin:
Thanks for catching this, and sorry for the delay! You were right that `fastapi run` needs the CLI installed. The guide has since been updated to `poetry add "fastapi[standard]"`, which includes fastapi-cli per FastAPI's recommended install, so closing this as covered. Appreciate the report!

PR #1743, opened 2024-08-09
unified billing docs

nbw:
### Summary of changes

Unified Billing (Aggregated Billing) documentation.

### Preview

screencapture-localhost-4567-docs-about-billing-2024-08-09-16_45_56

### DIFF
--- a/about/billing.html.markerb
+++ b/about/billing.html.markerb
@@ -91,7 +91,7 @@ You can create multiple Linked Organizations for a single Billing Organization.
 - Each organization has its own [private network](/docs/networking/private-networking/). Organizations are isolated from each other, which might
 be advantageous from a security standpoint.
 - Detaching a Linked Organization from its Billing Organization is not possible at this time. Instead we suggest creating a new organization and moving resources there.
-- Linking an existing Organization to a Billing Organization is not possible at this time. Instead, create a new Linked Organization and move resources there.
+- Linking an existing Organization to a Billing Organization is not possible at this time. Instead, create a new Linked Organization and [move resources](/docs/apps/move-app-org/) there.
 
 ---
 
--- a/about/billing.html.markerb
+++ b/about/billing.html.markerb
@@ -82,8 +82,9 @@ Unified billing allows you to consolidate billing for multiple organizations und
 
 ### How to use Unified Billing
 
-- A **Billing Organization** is just a normal organization on a plan. You can use an existing organization if you have one or create one from the [new organization page](/organizations/new).
-- A **Linked Organization** can be created from the [new organization page](/organizations/new).
+A **Linked Organization** can be created from the [new organization page](/organizations/new). When creating a new organization, you can choose an existing Organization to link it to.
+
+You can create multiple Linked Organizations for a single Billing Organization.
 
 ### Some things to consider
 
--- a/about/billing.html.markerb
+++ b/about/billing.html.markerb
@@ -12,15 +12,19 @@ The amount you're charged depends on your plan. For the Pay As You Go plan, you 
 
 Refer to our [Plans page](https://fly.io/plans) and our resource [Pricing page](/docs/about/pricing/) for details.
 
+---
+
 ## View invoices
 
 You can view and download invoices from the **Billing** section of your [dashboard](https://fly.io/dashboard/personal/billing) for each organization.
 
 To download a past invoice, click **View** for the relevant cycle. You'll be sent to the Stripe billing portal where you'll have the option to download the invoice.
 
+---
+
 ## Plan billing
 
-You're charged for Launch, Scale, and Enterprise plans as soon as you sign up for them. 
+You're charged for Launch, Scale, and Enterprise plans as soon as you sign up for them.
 
 New customers that are automatically signed up for the Pay As You Go plan are billed monthly for resource usage. For existing Fly.io accounts, when you create a new organization, you're signing up for the Pay As You Go plan.
 
@@ -32,6 +36,8 @@ Plan billing, and any included usage, is pro-rated over the month. For example, 
 
 When you upgrade to a higher plan, your billing is pro-rated and adjusted accordingly.
 
+---
+
 ## Machine billing
 
 Started Machines are billed per second that they're running (the time they spend in the `started` state), based on the price of a named CPU/RAM combination, plus the price of any additional RAM you specify.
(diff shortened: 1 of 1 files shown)


about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+
+- A **Billing Organization** is just a normal organization on a plan. You can use an existing organization if you have one or create one from the [new organization page](/organizations/new).
+- A **Linked Organization** can be created from the [new organization page](/organizations/new).
+
+### Some things to consider
+
+- Each organization has its own [private network](/docs/networking/private-networking/). Organizations are isolated from each other, which might
+be advantageous from a security standpoint.
+- Unlinking a Linked Organization from its Billing Organization is not possible at this time. Instead we suggest creating a new organization and moving resources there.
+- Linking an existing Organization to a Billing Organization is not possible at this time. Instead, create a new Linked Organization and moving resources there.
```
andie787, line 94:
```suggestion
- Linking an existing Organization to a Billing Organization is not possible at this time. Instead, create a new Linked Organization and move resources there.
```

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
 
 Volume billing is pro-rated to the hour and we subtract the free allowances first for the Launch, Scale, and Enterprise plans. For details, see [Volume pricing](/docs/about/pricing/#persistent-storage-volumes).
 
 If you create a volume, you will be charged for it. You’re billed for volumes that aren’t attached to Machines, and for volumes that are attached to Machines in any state, including stopped Machines.
 
+---
+
+## Unified Billing
+
+Unified billing allows you to consolidate billing for multiple organizations under a single organization. This means  You'll receive a single invoice for all organizations under unified billing. This feature is useful for managing billing for multiple organizations, such as when you have multiple teams or projects.
```
andie787, line 71:
```suggestion
Unified billing allows you to consolidate billing for multiple organizations under a single organization. This means you'll receive a single invoice for all organizations under unified billing. This feature is useful for managing billing for multiple organizations, such as when you have multiple teams or projects.
```

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+
+**Billing Organization**: a normal organization with a payment method linked to it. This organization is responsible for paying for the resources used by all organizations linked to it.
```
andie787, line 73:
```suggestion
**Billing Organization**: The organization that is responsible for paying for the resources used by all organizations linked to it. Any organization in good standing with an attached payment method can be a Billing Organization.
```

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+
+**Linked Organization**: an organization that has its billing linked to a another organization. It otherwise functions a normal organization
```
andie787, line 75:
```suggestion
**Linked Organization**: An organization that is linked to another another organization for billing purposes. It otherwise functions as a normal organization.
```

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+
+### How Unified Billing Works
+
+- You'll receive a single invoice for all organizations under unified billing
+- Billing Organization must have a payment method on file (credit card or Stripe Link)
+- Linked Organizations inherit the plan of their Billing Organization
```
andie787, line 81: > - Linked Organizations inherit the plan of their Billing Organization

this might raise questions. can we clarify what it means exactly?

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+- Linked Organizations share credits with their Billing Organization
+
+### How to use Unified Billing
+
```
andie787, line 85: I think we should make this more "how to".

Something using what you have but more like:

Choose an organization to be the Billing Organization. Any organization on a plan can become the Billing Organization. (Question: does this include Pay As You Go? What about legacy plans?) You can use an existing organization if you have one or create one from the [new organization page](/organizations/new) in the dashboard.

Create a new organization using the [dashboard](/organizations/new), where you'll have the option to link it to a Billing Organization. This new org will be a Linked Organization. You can create and link multiple organizations to a single Billing Organization.

(not happy with the last paragraph above, but we can rework it)

andie787 (commented):
sry if you weren't ready for review yet, but with timezones I thought I'd do it now so you have it Monday :)

about/billing.html.markerb
nbw: I actually think it might be more-than-necessary information. we can clarify this detail if customers start to ask, but it's probably a distinction that only matters to those of us making the platform.

about/billing.html.markerb
nbw: >  (Question: does this include Pay As You Go? What about legacy plans?)

any plan. I'd like for us to avoid talking about plans in at all with this feature, actually.

> Create a new Linked Organization using the [dashboard](https://github.com/organizations/new),

it's not actually from then dashboard. it's from the new organization page which isn't always obvious. that's why I was linking to that page specifically. 

---
@andie787 I've updated the copy. It's much simpler.

I decided lead with instructions on how to create a Linked Organization (since that is the "new" organization type that's being added by this feature).

Creating a billing organization is very obvious from the UI. The user won't have any organizations to select from in the dropdown if they don't have a billing organization (and there are instructions to get them to create one). If we start to get questions about "how do I create a billing organization" then I think we'll add it. I'm inclined to not confuse the user into thinking that there's a special thing called a "billing organization" when it's way more simple than that

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+A **Linked Organization** can be created from the [new organization page](/organizations/new). When creating a new organization, you can choose an existing Organization to link it to.
+
+You can create multiple Linked Organizations for a single Billing Organization.
+
+### Some things to consider
+
+- Each organization has its own [private network](/docs/networking/private-networking/). Organizations are isolated from each other, which might
+be advantageous from a security standpoint.
+- Detaching a Linked Organization from its Billing Organization is not possible at this time. Instead we suggest creating a new organization and moving resources there.
```
nbw, line 93: the linter really hates the word "Unlinking" so I used "Detaching" instead.

about/billing.html.markerb
andie787: This all makes sense. And I like/agree with not mentioning plans at all.

I think I end up calling anything in the UI the "dashboard". (TBD: We should probably standardize how we talk about these pages and how we do links in docs!)

about/billing.html.markerb
andie787: I added "unlinking" to the linter for you if you want to use it!

about/billing.html.markerb
```diff
@@ -42,22 +48,59 @@ Stopped Machines are billed based on their root file system (rootfs) usage per s
...
+- Linking an existing Organization to a Billing Organization is not possible at this time. Instead, create a new Linked Organization and move resources there.
```
andie787, line 94:
```suggestion
- Linking an existing Organization to a Billing Organization is not possible at this time. Instead, create a new Linked Organization and [move resources](/docs/apps/move-app-org/) there.
```

andie787 (commented):
Looking good! Left one suggestion and a couple comments. You're good to merge when you're ready.

btw, I forgot to mention that I really like the UI for this feature. It manages to be both clean and helpful.

PR #1744, opened 2024-08-09
Add a Tokens page to Security section

andie787:
### Summary of changes

Add a bare-bones page on the different tokens under security. This can be the starting point for a whole section on tokens. Let's call this the start of continuous improvements.

### Preview

https://aa-docs.fly.dev/docs/security/tokens/


security/tokens.html.markerb
```diff
@@ -0,0 +1,150 @@
...
+
+<div class="important icon">
+**Important:** We've previously suggested in docs and community that you should use the output of the `fly auth token` command as your API token for everything; this is no longer the best choice. Sometimes referred to as the "personal access token" or the "all-powerful auth token", this token is automatically created each time you log in with `fly auth login` and is used by flyctl to create, manage, configure, and deploy all the apps in any organization in your account. You don't need a token with that much scope in most cases, so we recommend using tokens created with `fly tokens deploy` or `fly tokens org` instead. It's always a good idea to use the token with the narrowest access that will work for your purpose.
```
btoews, line 13: It also might be worth mentioning that `fly auth token` tokens might expire unpredictably. People really shouldn't use them moving forward.

security/tokens.html.markerb
```diff
@@ -0,0 +1,150 @@
...
+
+You can create a standard org-scoped token or a read-only org-scoped token.
+
+## Create and manage tokens
+
+Create and manage tokens using flyctl. `fly tokens create` commands have some useful options for customizing your token:
+
+- **token names:** If you're managing multiple tokens, then give the tokens custom names with the `--name` option to make them easier to find when you list them with flyctl or view them in your dashboard.
+
+- **token expiry:** When you create a token, include the duration that the token is valid for with the `--expiry` option. In general, you should specify the shortest possible duration for your use case so that you limit access by time as well as scope. Tokens are valid for 175200h0m0s (20 years) by default.
```
btoews, line 36: > 175200h0m0s

We can probably just say "20 years"

security/tokens.html.markerb
```diff
@@ -0,0 +1,150 @@
...
+Create a read-only org-scoped deploy token with a custom name and expiry:
+
+```
+fly tokens create readonly --name <"my token name"> --expiry <duration>
+```
+
+Create a read-only token based on an existing org-scoped token:
+
+```
+fly tokens create readonly --name <"my token name"> --expiry <duration> --from-existing <token ID>
```
btoews, line 115: Does this work? I'd expect you to have to pass the existing token itself rather than it's ID.

btoews (approved):
I love it! Thank you for writing this!

security/tokens.html.markerb
andie787: I could not get this to work at first, since I was misunderstanding how the option worked (spoiler: `--from existing`  doesn't take the token as an argument!) I will update the text

PR #1750, opened 2024-08-10
Update autoscale-by-metric - fix typo in `secrets set` cli command

audacioustux:


andie787:
😱 thanks for this fix!

PR #1777, opened 2024-08-16
Clarify that suspended machines are billed the same as stopped

billyb2:

### DIFF
--- a/about/billing.html.markerb
+++ b/about/billing.html.markerb
@@ -44,7 +44,7 @@ Started Machines are billed per second that they're running (the time they spend
 
 For example, a Machine described in your dashboard as "shared-1x-cpu@1024MB" is the “shared-cpu-1x” Machine size preset, which comes with 256MB RAM, plus additional RAM (1024MB - 256MB = 768MB). For pricing and available CPU/RAM combinations, see [Compute pricing](/docs/about/pricing/#compute).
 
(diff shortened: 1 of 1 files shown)


about/billing.html.markerb
```diff
@@ -44,7 +44,7 @@ Started Machines are billed per second that they're running (the time they spend
 
 For example, a Machine described in your dashboard as "shared-1x-cpu@1024MB" is the “shared-cpu-1x” Machine size preset, which comes with 256MB RAM, plus additional RAM (1024MB - 256MB = 768MB). For pricing and available CPU/RAM combinations, see [Compute pricing](/docs/about/pricing/#compute).
 
-Stopped Machines are billed based on their root file system (rootfs) usage per second (the time they spend in the `stopped` state) by $0.15 per GB per month.
+Stopped Machines are billed based on their root file system (rootfs) usage per second (the time they spend in the `stopped` state) by $0.15 per GB per month. The same goes for `suspended` machines.
```
andie787, line 47:
```suggestion
Stopped and suspended Machines are billed based on their root file system (rootfs) usage per second (the time they spend in the `stopped` or `suspended` state) by $0.15 per GB per month.
```

andie787 (commented):
thanks for noticing that we missed this!
suggesting a change for clarification and feel free to merge when your happy with it

Issue #1772, opened 2024-08-15
Issue with the "Backup, Restores, & Snapshots" doc -- Snapshot Restores


swilson-fly:
I found an issue with this document.

Title: Backup, Restores, & Snapshots
Location: https://fly.io/docs/postgres/managing/backup-and-restore/
Source: https://github.com/superfly/docs/blob/main/postgres/managing/backup-and-restore.html.markerb

### Describe the issue

There's some interop issues with restoring a snapshot for an older stolon PG into a new postgres-flex cluster. It's possible there's issues restoring to diff pg-flex major versions as well! Restoring to the same image is best, but looks like there's some pg-flex versions that are more compatible than others. Ref: https://community.fly.io/t/reuse-old-volume-in-new-machine-after-irrepairable-hardware-damage/21230/34

### Addtional info

Todo: add guidance for specifying images.

AshitaGK:
Hi @swilson-fly,
By restoring a snapshot to the new pg-cluster based on the pg-flex versions, do you mean appending `--image-ref` to fly postgres create --snapshot-id <snapshot-id>
command?

swilson-fly:
@AshitaGK Yes, exactly! By default `fly postgres create` makes a new cluster running postgres-flex:16. For snapshots from clusters running older versions this may not be compatible, so users need to specify the same image as their original cluster. 

PR for a doc fix is here and will be shipped today https://github.com/superfly/docs/pull/1799

PR #1799, opened 2024-08-22
Update PG Snapshot restore instructions

swilson-fly:
### Summary of changes

There's some interop issues with restoring a snapshot for an older stolon PG into a new postgres-flex cluster. It's also not ideal to restore across different major versions of pg-flex. This updates the PG restore docs to add a section on  how to find your image and specify it when restoring the pg cluster. It also updates the PG restore instructions in the single host failure doc.

Resolves https://github.com/superfly/docs/issues/1772

### DIFF
--- a/postgres/managing/backup-and-restore.html.markerb
+++ b/postgres/managing/backup-and-restore.html.markerb
@@ -84,12 +84,25 @@ vs_OPQXXna6kA2Qnhz8 26 MiB 2 days ago
 
 The values under the `ID` columns are what will be used to restore a snapshot.
 
-## Restoring from a snapshot
-
-To restore a Postgres application from a snapshot, simply specify the `--snapshot-id` argument when running the `create` command as shown below:
+## Identifying your Postgres image version
+Depending on when you created your Postgres cluster, it may be running an older image than the default for newly clusters. Different Postgres major versions may not be fully compatible, so it's important to use the same version for your restored cluster.
+To see your Postgres image and version, run `fly image show`. 
 
 ```cmd
-fly postgres create --snapshot-id <snapshot-id>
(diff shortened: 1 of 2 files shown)


postgres/managing/backup-and-restore.html.markerb
```diff
@@ -84,12 +84,27 @@ vs_OPQXXna6kA2Qnhz8 26 MiB 2 days ago
...
+MACHINE ID      REGISTRY                REPOSITORY      TAG     VERSION DIGEST                        LABELS                                                                          
+e286004f696700  registry-1.docker.io    flyio/postgres  14.6    v0.0.41 sha256:3c25db96357a78e827ca7d fly.app_role=postgres_clusterfly.pg-version=14.6-1.pgdg110+1fly.version=v0.0.41
+```
+The values under the `REPOSITORY and `TAG columns are the image you'll use to restore the snapshot. 
+Legacy postgres images use the `flyio/postgres` repository, while new Postgres Flex images use the flyio/`postgres-flex` repository. In the above example, the machine is running a legacy `flyio/postgres:14.6` image.
+
 ## Restoring from a snapshot
 
-To restore a Postgres application from a snapshot, simply specify the `--snapshot-id` argument when running the `create` command as shown below:
+To restore a Postgres application from a snapshot, simply specify the `--snapshot-id` argument and the `--image-ref` argument when running the `create` command as shown below:
```
Roadmaster, line 104: One has to also specify `--stolon` if it's a Stolon-based cluster (14.x and lower, I think). For a few people who tried a restore, even with the correct image,  without `--stolon` replication is not set up correctly. Running with `--stolon` fixes things.

PR #1816, opened 2024-08-29
fully write out powershell commands

h8ku:
If you have Elixir installed, using the short version of the command `iex` breaks the install. Better to fully write it out. People will copy and paste it anyway.


kcmartin:
Thanks for this, and sorry for the long delay! You're right that a re-aliased `iex` (common with Elixir installed) can break the piped install, and spelled-out cmdlet names would avoid that. But this change also swaps `pwsh` for legacy `powershell`, which contradicts the recommendation right below the command to prefer modern PowerShell. Closing this one, but if you're still up for it, we'd happily take a new PR that spells out the cmdlet names while keeping `pwsh`.

Issue #1798, opened 2024-08-22
Minor correction in Rust framework guide


AshitaGK:
Title: Axum repeated in other Rust frameworks for installation.

Source: https://fly.io/docs/rust/frameworks/actix-web/

Issue: Minor correction but it seems to be repeated in every other Rust framework. The line 'If you don’t already have an existing Axum application' needs to be modified according to the rust framework you want to install. 
![Screenshot_20240822-211243](https://github.com/user-attachments/assets/23466129-29b8-4541-960d-901af1f40c4b)

PR #1822, opened 2024-09-05
Update _cargo-new.html.erb

AshitaGK:
### Summary of changes
Dynamically insert the correct framework name based on the runtime variable passed in the locals object when rendering the partial.
The documentation will reference the correct framework instead of Axum as a default framework.


AshitaGK:
@andie787 for reference, please check this issue->https://github.com/superfly/docs/issues/1798

PR #1828, opened 2024-09-20
fix: add health check to exporting logs documentation

rolznz:
### Summary of changes

The log shipper sometimes stops working randomly, so it's good to have a health check. This configuration was recommended to me by Leslie from Fly support.

### Preview

N/A

N/A

### Notes

N/A


kcmartin:
Sorry for the long silence on this, and thanks for flagging the silent-failure problem. Since this was opened we've reworked the page in #2420 to remove the public service entirely (it exposed Vector's API publicly), so this diff no longer applies. We've added a health check section there instead, using a top-level `[checks]` block with your check settings, which works without a public service. Closing in favor of #2420. Thanks again for the contribution!

PR #1843, opened 2024-10-17
Fix PostgreSQL fact in index.html.md

amingilani:
PostgreSQL is officially still named PostgreSQL, but Postgres is an accepted nickname for the project.

### Summary of changes

This patch fixes the portion of the docs that call PosgreSQL a former name of Postgres on the [Fly Postgres docs](https://fly.io/docs/postgres/#fully-managed-postgres)


amingilani:
Whoops, I think I misread the sentence. Please ignore, I'm sick today.

Issue #1844, opened 2024-10-17
Issue with the "Setting up a Python Environment" doc


mattt:
I found an issue with this document.

Title: Setting up a Python Environment
Location: https://fly.io/docs/python/the-basics/initial-setup/
Source: https://github.com/superfly/docs/blob/main/python/the-basics/initial-setup.html.md

### Describe the issue

The guide for setting up a Python environment should mention [uv](https://docs.astral.sh/uv/) as an option for dependency management.

Especially since its 0.3 release, uv has become a [fully-fledged packaging tool](https://astral.sh/blog/uv-unified-python-packaging). So beyond mentioning it as an alternative, I think it's worth considering updating the guide making uv the new default, replacing Poetry.

mattt:
I opened a PR with these changes here: https://github.com/superfly/docs/pull/1847

kcmartin:
Thanks for raising this, and for the earlier PR. We've added uv to the list of dependency-management options in the Python setup guide in #2457, so it's now called out as a supported choice.

We're holding off on the larger change of making uv the default over Poetry for now, since our Python example repos and framework guides are still built around Poetry, and switching the default would mean updating those in lockstep to keep the docs consistent. Wanted to at least get uv mentioned in the meantime. This will close when #2457 merges.

PR #1847, opened 2024-10-19
Update Python docs to use `uv` instead of `poetry`

mattt:
### Summary of changes

This PR updates the Python guides to use `uv` for dependency management.

https://github.com/superfly/docs/issues/1844

### Notes

- The ["Setting up a Python Environment" guide](https://github.com/superfly/docs/pull/1847/files#diff-2873f777ffeaf6a2b6a1f7e50a8533251fa45a176f4e6867b528efd163d84f7b) is mostly the same. Because `uv` uses venv by default, fewer configuration steps are required.
- The ["Multi-stage Builds" guide](https://github.com/superfly/docs/pull/1847/files#diff-4c97ace71199bd256d3bdf2b3581229dc6bfe1edaa4b5e05f5643aef658fd292) is a more involved change.
- The `poetry_new` partial is replaced with a new [`uv_init` partial](https://github.com/superfly/docs/pull/1847/files#diff-02947418b2c18317c04d287be133d5bac7b31b285c9c6196892c149e792ede96), and all of its references are updated.
- The ["Getting Started"] guide for Django still uses `pip` + `venv`, and is updated only to mention `uv` as an alternative (as it currently does for Poetry)
- The remaining changes are simple replacements of `poetry add` → `uv add`, which does the same thing.

### DIFF
--- a/django/getting-started/index.html.md
+++ b/django/getting-started/index.html.md
@@ -34,7 +34,11 @@ mkdir hello-django && cd hello-django
 
 ### Virtual Environment
 
-For this guide, we use [venv](https://docs.python.org/3/library/venv.html#module-venv) but any of the other popular choices such as [Poetry](https://python-poetry.org/), [Pipenv](https://github.com/pypa/pipenv), or [pyenv](https://github.com/pyenv/pyenv) work too.
+For this guide, we use [venv](https://docs.python.org/3/library/venv.html#module-venv) but any of the other popular choices such as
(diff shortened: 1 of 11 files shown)


kcmartin:
Sorry for the long silence here, and thanks for the thorough work. The direction is right, but we don't have the capacity right now to switch the recommended tooling across the Python docs and support it properly. Closing this for now, with #1844 open to track making uv the default when we can.

PR #1862, opened 2024-10-31
Improve http_service.checks documentation

scottohara:
### Summary of changes
For the `method` and `protocol` options, indicate that the default values (if omitted) are `get` and `http`, respectively.

Additionally, add a warning section that warns about common issues with redirects (or any non-200 response).

One such example is that, by default, a standard Rails application includes `config.force_ssl = true` in `config/environments/production.rb`, which causes a health check request to http://0.0.0.0:3000/up to receive a 301 redirect to https://0.0.0.0:3000/up (and the default puma config for Rails is not configured to accept SSL requests).

The warning includes some guidance about adding an `X-Forwarded-Proto = "https"` header to the health check to avoid this sort of redirect.

https://community.fly.io/t/http-service-checks-random-questions/22424


scottohara:
Just bumping this low priority PR as it may have slipped off everybody’s radar. 

There was initially some linter errors on the first build attempt, but I believe they have been addressed so this seems to just be awaiting an approval to rerun the workflow.

Thanks.

PR #1886, opened 2024-12-17
update attached volume limitation on canary and bluegreen strategies

kcmartin:
### Summary of changes

A user reported an error when trying to use the `canary` deployment strategy on a Machine with an attached volume. This limitation was confirmed for both canary and bluegreen strategies. Updating the docs to clarify and reflect this limitation.


Roadmaster (approved):
LGTM! Thanks.

PR #1897, opened 2024-12-30
fix: Make header level of "Fixed price plans" match "Pay-as-you-go plan"

lukasgjetting:
### Summary of changes

- Header for the two plan types now match, previously they didn't

### Preview

Before:

Screenshot 2024-12-30 at 20 29 32


kcmartin (approved):
Thanks for fixing the header level

PR #1909, opened 2025-01-15
Update runtime-environment.html.md

thomasfr:
Clarified when its possible to set `--region` flag. 

### Summary of changes
Currently the docs for "PRIMARY_REGION" says "during deploys". This PR tries to clarify it by replacing it with "when using `fly launch`".


kcmartin (approved):
LGTM, sorry for the long delay!

PR #1910, opened 2025-01-16
Add a warning message to min_machines_running

Lucais11:
### Summary of changes
Added a warning description under min_machines_running for users, to make sure it is clear that the setting only maintains the specified minimum number of running machines in the app's primary region


Lucais11:
The linter doesn't like the word "fly.io" in the url not being capitalized and the word "machines" in min_machines_running not being capitalized. Both of which wouldn't make sense to capitalize in this sense.

PR #1911, opened 2025-01-17
Update machine restart policy docs

wjordan:
### Summary of changes

Add extra details to the `on-fail` restart policy documentation:
- clarify that the `max_restart` value is configurable, with a default 10
- add detail that the max number of restarts is evaluated within a fixed 5-minute window

https://community.fly.io/t/one-of-my-apps-both-machines-just-stopped/23488/4?u=wjordan
https://github.com/superfly/nomad-firecracker/pull/1813


kcmartin:
Added a couple of commits to fix Vale linter error and a field name. Sorry for the long delay.
LGTM, merging in

PR #1915, opened 2025-01-24
Fix typo (missing 'is') in secrets.html.markerb

thad-imbue:


kcmartin (approved):
Thanks for the typo fix

PR #1916, opened 2025-01-26
removed dot from mysql example name since it fails

mabis:
using a dot in the app name as suggested results in an errors (both console and web) since dots are not allowed (anymore, I guess)

"app name must consist of only lowercase letters, numbers, and dashes"


mabis:
my edit was not complete, the text needs more reworking to make it clear to the user that the name cannot contain dot but the, .internal will be automatically added and used when they will connect to the mysql server from another machine

PR #1917, opened 2025-01-26
Update mysql-on-fly.html.markerb to use at least mysql:8.0.37

mabis:
as reported also elsewhere (e.g. https://community.fly.io/t/error-deploying-mysql-database/20008/4 ) the mysql:8 image suggested in the example fails to work in the end (mysqld keeps erroring out due to missing mysql.component and unknown default auth plugin).

This is quickly fixed by using at least v 8.0.37. 

Haven't checked other versions, this is good enough for anyone taking this for a quick spin


kcmartin (approved):
Thanks, this looks good for now

PR #1924, opened 2025-01-31
Fix links in LiteFS docs

ranok:
### Summary of changes
Fix the links in the LifeFS docs index. It looks like there was some variable setting that was not working for any of the links on this page, so they've been set as explicit links.

- https://fly.io/docs/litefs/


ranok:
Duplicate of #1914

PR #1947, opened 2025-02-18
fix links

onaralili:
### Summary of changes
Fixes broken links under the https://fly.io/docs/litefs/#exploring-our-guides

### DIFF
--- a/litefs/index.html.markerb
+++ b/litefs/index.html.markerb
@@ -28,12 +28,12 @@ You can get up and running quickly with one of our guides:
 
 - [Speedrun: Adding LiteFS to your app](/docs/litefs/speedrun) the fastest way to get started with LiteFS on Fly.io.
 
-- [Getting Started on Fly.io][] helps you add LiteFS to an existing application and deploy to Fly.io. This guide
+- [Getting Started on Fly.io](/docs/litefs/getting-started-fly/) helps you add LiteFS to an existing application and deploy to Fly.io. This guide
 provides more details and explanation than the Speedrun.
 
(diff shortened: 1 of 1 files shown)


litefs/index.html.markerb
```diff
@@ -28,14 +28,14 @@ You can get up and running quickly with one of our guides:
 
 - [Speedrun: Adding LiteFS to your app](/docs/litefs/speedrun) the fastest way to get started with LiteFS on Fly.io.
 
-- [Getting Started on Fly.io][] helps you add LiteFS to an existing application and deploy to Fly.io. This guide
+- [Getting Started on Fly.io](/docs/litefs/getting-started-fly/) helps you add LiteFS to an existing application and deploy to Fly.io. This guide
 provides more details and explanation than the Speedrun.
 
-- [Getting Started with Docker][] helps you add LiteFS to an existing application that you want to run outside of Fly.io.
+- [Getting Started with Docker][/docs/litefs/getting-started-docker/] helps you add LiteFS to an existing application that you want to run outside of Fly.io.
```
kittsville, line 34:
```suggestion
- [Getting Started with Docker](/docs/litefs/getting-started-docker/) helps you add LiteFS to an existing application that you want to run outside of Fly.io.
```

Two pairs of square brackets in markdown is used for [reference links](https://www.markdownguide.org/basic-syntax/#reference-style-links), where the second block of text is not a URL but a reference to later in the page where the URL is defined. If you want to provide the link inline then the second pair of brackets must be regular parenthesis:

```
Just this:
[link text](https://example.com)
Or these two instead:
[link text][reference-id]
[reference-id]: https://example.com
```

kcmartin:
Duplicate of #1914

PR #1954, opened 2025-02-24
Fix LiteFS links

tombailey:
### Summary of changes

Fixes broken links on https://fly.io/docs/litefs/

### Notes

Not sure what the intention of `[Using LiteFS Cloud for Backups]: /docs/litefs/cloud-backups` was. Perhaps it is unused?

PR #1958, opened 2025-03-08
Update index.html.markerb

nunoskew:
Fixed following links on the bottom of https://fly.io/docs/litefs/:

Getting Started with LiteFS on Fly.io
Getting Started with LiteFS in Docker
How LiteFS Works


nunoskew:
Duplicate of #1954

PR #1959, opened 2025-03-08
Update ruby.html.markerb

jasonwbarnett:
fix typo


kcmartin (approved):
Thanks for the typo fix

PR #1962, opened 2025-03-14
Removed deprecated command from the docs and clarified required format for the --image-ref argument 

eporto2021:
### Summary of changes

Removed deprecated command from the docs and clarified required format for the --image-ref argument in the ```fly postgres create --snapshot-id <snapshot-id>  --image-ref $REPOSITORY:$TAG``` command.


kcmartin (approved):
Sorry for the long wait on this one. You were right on both counts: `fly info` is deprecated and the `--image-ref` format needed clarifying. I pushed a small cleanup commit on top of your change to fix a couple of typos and match our docs placeholder style. Approving and merging now, thanks for the contribution!

PR #1990, opened 2025-04-04
Remove reference to Tigris data transfer being free

injoongy:
### Summary of changes
- Removed line about data transfer to Tigris being free.
- Removed references to the `-o` flag for the `fly tokens create` command, which is not a valid flag.

- [Community forum post](https://community.fly.io/t/tigris-billing-for-inbound-data-from-fly-tigris/24164) about Tigris data transfer
- [Internal Discourse discussion](https://flyio.discourse.team/t/customer-getting-billed-for-data-transfer-into-tigris/8107) about Tigris data transfer pricing
- [HS ticket](https://secure.helpscout.net/conversation/2865475001/124530?viewId=6664451) for the `-o` flag

### DIFF
--- a/launch/autoscale-by-metric.html.markerb
+++ b/launch/autoscale-by-metric.html.markerb
@@ -44,10 +44,10 @@ fly secrets set -a my-autoscaler --stage FAS_API_TOKEN="FlyV1 ..."
 The next auth token you'll need is one that has permissions to read from your organization's Prometheus data on Fly:
 
 ```cmd
-fly tokens create readonly
+fly tokens create readonly my-org
 ```
 
-You'll be prompted to choose which org you want to create it for. Afterwards, copy the token and use it as a secret on your autoscaler app:
+Copy the token and use it as a secret on your autoscaler app:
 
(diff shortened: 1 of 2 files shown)


launch/autoscale-by-metric.html.markerb
```diff
@@ -44,10 +44,10 @@ fly secrets set -a my-autoscaler --stage FAS_API_TOKEN="FlyV1 ..."
 The next auth token you'll need is one that has permissions to read from your organization's Prometheus data on Fly:
 
 ```cmd
-fly tokens create readonly -o my-org
```
jphenow, line 47: Does specifying an org without -o work? I suspect we were trying to avoid forcing people into an interactive CLI in case they wanted to use this for scripts that won't have interactive shell.

jphenow (approved):

launch/autoscale-by-metric.html.markerb
injoongy: Ahh, it does - let me update this to add that, thanks!

PR #2000, opened 2025-04-13
Use four slases for absolute paths in sqlite database connection string

deepakjois:


kcmartin:
Sorry for the late reply, and thanks for the PR. Four slashes is the SQLAlchemy convention; in Rails, `sqlite3:///mnt/name/production.sqlite` already resolves to the absolute path `/mnt/name/production.sqlite`, so the current URL is correct. Closing, but thanks for the contribution.

PR #2020, opened 2025-04-29
clarify required settings for volume auto-extension

Lucais11:
### Summary of changes
Added a callout to clarify that both auto_extend_size_increment and auto_extend_size_limit must be set for the volume auto-extension feature to work. Removed individual “Required with...” lines to reduce redundancy and improve clarity.


Lucais11:
Doesn't look like the failed Vale linter applies to my PR. Its unrelated to what I've changed

PR #2024, opened 2025-04-30
rewrite Machine states page with full lifecycle and versioning detail

Lucais11:
### Summary of changes
This update overhauls the Machine states documentation to provide a clearer, more complete view of how Fly Machines transition through lifecycle states. The updated page:

Differentiates between overall Machine state and version-specific state

Categorizes states as persistent, transient, or terminal, with full descriptions

Adds detail on how skip_launch affects Machine updates and startup behavior

Covers diagnostics for Machines stuck in transitional states

Introduces two Mermaid diagrams to visualize state transitions

Clarifies behavior of replaced and migrated terminal states

Adds related links to supporting docs for context

The new structure better supports users automating Machine orchestration or interacting with the Machines API directly.


swilson-fly (approved):
👍

PR #2027, opened 2025-05-05
clarify volume create region behavior in Machines API docs

Lucais11:
### Summary of changes
Clarified the region field behavior in the Machines API "Create a volume" docs. Updated the description to reflect that region is optional when attaching a volume to a Machine, and that it defaults to the Machine’s region or the app’s primary region if not specified.


Lucais11:
The vale linter errors don't apply to my changes that were made.

PR #2034, opened 2025-05-08
clarify Fly Proxy behavior when autostart/autostop are not set

Lucais11:
### Summary of changes
Added a new section explaining how Fly Proxy behaves when neither `auto_start_machines` nor `auto_stop_machines` is set in the `fly.toml` file.


Lucais11:
The Vale linter isn't in regards to my changes

PR #2086, opened 2025-06-16
Update support mpg note

Lucais11:
### Summary of changes
Add MPG to the supported products column and change the note at the very bottom of the page to reflect this update


Lucais11:
Vale linter is freaking out because of the imbedded link. Doesn't apply

PR #2099, opened 2025-06-26
Update app-services page to include network-debugging tip

max-flint:
### Summary of changes
Some environments still run IPv4 only. If a user only assigns an IPv6 address to their service, it be difficult to debug why a typical `curl https://<app>.fly.dev` command does not work.

Adds a debugging tip for users in these network situations.


kcmartin (approved):
Thanks @max-flint, good catch. I pushed a small edit to make the tip actionable (check fly ips list, plus the shared IPv4 command and doc link), and adding some terms to the Vale linter vocab. Merging now.

PR #2162, opened 2025-07-31
add mpg regions coming soon

kcmartin:
### Summary of changes
- Add MPG regions that are coming soon (we can move them to the main list when they're actually available)
- Added US state to be consistent with our main regions list doc

### Preview
Previewed locally


thelukewhiting:
@kcmartin ship it!

PR #2173, opened 2025-08-09
Update using-symfony.html.md

dpi:
### Summary of changes

Examples need a little more code context to be correct.


kcmartin:
Thanks @dpi, both fixes were right. I added the missing `use Symfony\Component\Runtime\SymfonyRuntime;` on top so the class resolves under the new namespace, plus a couple of lint nits. Merging now.

PR #2179, opened 2025-08-13
Add best-effort warning about kill_timeout

kcmartin:
### Summary of changes
Add warning that the `kill_timeout` setting should be considered best-effort. There are times, for example, that it can interfere with required infra operations like host drain, so the customer's app needs to be able to handle that.

### Preview
Previewed in local env


injoongy (approved):
LGTM - thanks for adding this!

kcmartin:
Verified changes are live - https://fly.io/docs/reference/configuration/#kill_timeout-option

PR #2192, opened 2025-08-21
Update configuration.html.markerb

mstahv:
### Summary of changes
Changed to a builder that is still maintained.

Not sure if the separate buildpacks parameter is needed, but kept that as as separate example.

### Notes

I wonder why you don't fall back to buildpacks if Dockerfile does not exist🤷‍♂️ At least it could do a priming build if it succeeds and if so, then drop this line to the tly.toml. Tried your platform first time today and experience was not very good. With your docs and google I expected things would work and then from docs I got snippet that brought me outdated buildpack. Implicitly using up-to-date buildpacks might make me try this again (or promote this service to our developer audience)

### DIFF
--- a/reference/configuration.html.markerb
+++ b/reference/configuration.html.markerb
@@ -98,13 +98,19 @@ The optional build section contains key/values concerned with how the applicatio
 
 ```toml
 [build]
-  builder = "paketobuildpacks/builder:base"
-  buildpacks = ["gcr.io/paketo-buildpacks/nodejs"]
+  builder = "paketobuildpacks/builder-jammy-base"
 ```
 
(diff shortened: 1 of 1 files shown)


kcmartin:
Hi @mstahv, sorry for the long delay on this one. You're right that `paketobuildpacks/builder:base` is no longer maintained, thanks for the switch to the correct one. I pushed a small commit to your branch updating the second example as well. Thanks for the contribution!

The failing Vale check isn't from this PR. The linter flags pre-existing term-casing on words inside code (like [[vm]]), which can't be changed. It's non-blocking and I'm merging this one. We've also disabled that noisy rule in the Vale config (#2432, merged, current Vale failure is just a stale run against the old config and will clear on the next fresh CI run).

PR #2205, opened 2025-09-03
Add tokens clarification for authorization headers

Lucais11:
### Summary of changes
Clarified which tokens use Bearer vs FlyV1 in the metrics docs.

### DIFF
--- a/monitoring/metrics.html.md
+++ b/monitoring/metrics.html.md
@@ -60,7 +60,7 @@ Queries can be sent to the following endpoint:
 https://api.fly.io/prometheus/<org-slug>/
 ```
 
-You’ll need to authenticate with a Fly Access Token. Depending on the token type, the header uses either the Bearer format (`Authorization: Bearer <token>`) or the FlyV1 format (`Authorization: FlyV1 <token>`). You may only query series scoped to your organizations.
+You’ll need to authenticate with a [Fly Access Token](https://fly.io/docs/security/tokens/). Depending on the token type, the header uses either the Bearer format (`Authorization: Bearer <token>`) or the FlyV1 format (`Authorization: FlyV1 <token>`). You may only query series scoped to your organizations.
 
 
 #### Manually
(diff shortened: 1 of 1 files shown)


monitoring/metrics.html.md
```diff
@@ -60,7 +60,8 @@ Queries can be sent to the following endpoint:
 https://api.fly.io/prometheus/<org-slug>/
 ```
 
-You'll need to authenticate with a Fly Access Token sent in the [standard](https://www.rfc-editor.org/rfc/rfc6750.html) Bearer Token format (e.g., an HTTP request header `Authorization: FlyV1 <TOKEN>`), and you may only query series scoped to your organizations.
+You’ll need to authenticate with a Fly Access Token. Depending on the token type, the header uses either the Bearer format (`Authorization: Bearer <token>`) or the FlyV1 format (`Authorization: FlyV1 <token>`). You may only query series scoped to your organizations.
```
kcmartin, line 63: It would be helpful to link to either: the section below 'using tokens" or another docs page that covers Fly Access tokens in more detail (if we have it).

monitoring/metrics.html.md
```diff
@@ -416,21 +419,24 @@ This kind of token or "[macaroon](https://fly.io/blog/macaroons-escalated-quickl
...
 These tokens look like this once generated: `FlyV1 fm2_lJPECAAAAAAAAC7txBAzYI6PRWhHLT...(a lot of base64-encoded text)`.
 
 #### Using tokens
-To use one of these tokens in an HTTP header, the "FlyV1" identifier replaces the "Bearer" token identifier. So it would look like this:
+Use the correct scheme based on token type:
 
+- `Bearer <token>` → for `flyctl auth token`.
+- `FlyV1 <token>` → for all tokens created with `fly tokens create`.
 
+Example with a token created via `fly tokens create readonly`:
```
kcmartin, line 439: The other mention in L437 of `fly tokens create` doesn't mention `readonly` -- it should probably be consistent (unless it's required in your example, in that case, you could modify L437 to list both (`fly tokens create` and `fly tokens create readonly`

kcmartin (commented):
Left two comments, we're good after those are addressed

kcmartin (approved):
Thanks! LGTM!

PR #2207, opened 2025-09-04
show how to suspend your own machine

rubys:
This isn't obvious, and should be mentioned.

### DIFF
--- a/reference/suspend-resume.html.md
+++ b/reference/suspend-resume.html.md
@@ -70,7 +70,7 @@ GET /v1/apps/{app_name}/machines/{machine_id}/wait?state=suspended
 POST /v1/apps/{app_name}/machines/{machine_id}/start
 ```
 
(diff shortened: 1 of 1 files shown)


reference/suspend-resume.html.md
```diff
@@ -70,6 +70,13 @@ GET /v1/apps/{app_name}/machines/{machine_id}/wait?state=suspended
 POST /v1/apps/{app_name}/machines/{machine_id}/start
 ```
 
+While usage of the Machines API in general requires an API token, suspending your own machine using the `/.fly/api` Unix socket directly does not:
```
kcmartin, line 73:
```suggestion
Generally, you need an API token to use the Machines API. But if you're just suspending _your own_ machine, you can skip the token and hit the `/.fly/api` Unix socket directly:
```

kcmartin (commented):
I left a copy edit suggestion -- otherwise LGTM

PR #2223, opened 2025-09-15
Update runtime options config reference with shutdown sequence behavior

kcmartin:
### Summary of changes
- Update the `kill_signal` and `kill_timeout` runtime options config reference with info on shutdown sequence behavior
- Copy edit to the rest of the sub-sections for clarity
- Inspired by info from the Solutions architecture guide

### Preview

Screenshot 2025-09-15 at 4 17 15 PM

### DIFF
--- a/reference/configuration.html.markerb
+++ b/reference/configuration.html.markerb
@@ -52,7 +52,7 @@ The following options are available to control the lifecycle of a running applic
 
 ### `kill_signal` option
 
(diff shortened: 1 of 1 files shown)


reference/configuration.html.markerb
```diff
@@ -70,6 +70,23 @@ kill_timeout = 120
 
 <section class="warning">Note: `kill_timeout` settings should be considered best-effort, and your app needs to be prepared to handle shorter stopping times</section>
 
+**How the shutdown sequence works**
+These options come into play during controlled shutdowns such as `fly deploy` or when a machine stops due to `auto_stop_machines`. The sequence looks like this:
```
Roadmaster, line 74: `fly machine stop` also does a shutdown sending the two signals and respecting kill_timeout.

Roadmaster (approved):
Nice! Just a small nitpick that the controlled shutdown sequence also applies on fly machine stop - I confirmed this experimentally :D

PR #2226, opened 2025-09-18
dd more detail to destroy volumes section

kcmartin:

### DIFF
--- a/volumes/volume-manage.html.markerb
+++ b/volumes/volume-manage.html.markerb
@@ -108,6 +108,8 @@ Create an exact copy of a volume, including its data. By default, we place the n
 
 For options, refer to the [`fly volumes fork` docs](/docs/flyctl/volumes-fork/) or run `fly volumes fork --help`.
 
+A detailed guide to [volume forking](/docs/blueprints/volume-forking/) is also available.
+
 ## Add a volume to an unmanaged Machine
 
 For Machines that aren't managed with Fly Launch (`fly.toml` and `fly deploy`), you can create a volume and attach it when you clone a Machine. You can also [clone a Machine with a volume](#clone-a-machine-with-a-volume) to get a new Machine with an empty volume.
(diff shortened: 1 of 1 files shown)


volumes/volume-manage.html.markerb
```diff
@@ -176,19 +178,29 @@ For options, refer to the [`fly machine clone` docs](/docs/flyctl/machine-clone/
...
 
 ```cmd
 fly volumes destroy <volume id>
 ```
 
 For options, refer to the [`fly volumes destroy` docs](/docs/flyctl/volumes-destroy/) or run `fly volumes destroy --help`.
 
+Notes:
+
+* If the volume is still mounted to a Machine, destroy might fail. Stop the Machine first, or unmount the volume.
```
kcmartin, line 200: @Roadmaster Do you think we need more detail about unmounting here?

Roadmaster: Unmounting the volume involves removing the `[mounts]` section in the configuration, but note this will result in _all_ machines in the app becoming unlinked to their volumes (since updating the configuration requires a redeploy anyway). So I don't think that's something we should recommend, or if we do, mention that it involves removing volumes from all machines.

Even if you stop the machine first, the volume cannot be destroyed directly: `Error: failed destroying volume: failed to destroy volume vol_rky3eeelzdlwjww4: failed_precondition: volume is currently bound to machine: 784eee3f40130`

The correct sequence would be to entirely destroy the machine, then nuke the volume. (this is OK because by definition the Fly machine itself doesn't contain any persistable/important data). Then, creating a new machine will also require creating a new volume for it. Options: fly machine clone or fly scale count take care of creating a new, empty volume to go with a new machine.

So: actually getting rid of a volume that's attached to a machine is known to be rather finicky :/

Let me know if you think these instructions belong here in the volume management page, I can help write something up.

volumes/volume-manage.html.markerb
```diff
@@ -176,19 +178,34 @@ For options, refer to the [`fly machine clone` docs](/docs/flyctl/machine-clone/
...
+
+2. Destroy the volume using its ID:
 
 ```cmd
 fly volumes destroy <volume id>
 ```
 
 For options, refer to the [`fly volumes destroy` docs](/docs/flyctl/volumes-destroy/) or run `fly volumes destroy --help`.
 
+**Volumes attached to Machines**
```
kcmartin, line 198: I've modified your suggestion a bit, let me know if this works!

Roadmaster: Looks perfect, thanks!

volumes/volume-manage.html.markerb
```diff
@@ -176,19 +178,34 @@ For options, refer to the [`fly machine clone` docs](/docs/flyctl/machine-clone/
...
+
+Volumes attached to Machines can’t be destroyed. You can check which Machine (if any) a volume is attached to by running `fly volumes list` First, stop and destroy the Machine, then the unattached volume can be deleted.
+
+If you’re replacing the Machine, use `fly scale count` or `fly machine clone` to launch a new one. This will create and mount a fresh volume based on your `fly.toml` config.
+
+To stop using volumes entirely, remove any `[mounts]` from `fly.toml`, redeploy to detach them, and then delete the volumes.
```
Roadmaster, line 204: nitpick: a couple of places say "delete" the volumes, do we want to standardize on calling it "destroy" the volumes? (also "deleted -> destroyed" a few lines above).

Roadmaster (approved):
LGTM with one small optional nitpick.

volumes/volume-manage.html.markerb
kcmartin: Thanks! Fixed

PR #2238, opened 2025-09-24
Add note on web-sockets for fly-replay

symkat:
### Summary of changes

Adding a note for web-sockets to clarify the router-app shouldn't negotiate web socket connections itself.

### DIFF
--- a/networking/dynamic-request-routing.html.markerb
+++ b/networking/dynamic-request-routing.html.markerb
@@ -241,6 +241,10 @@ If you replay with `prefer_instance` set, Fly Proxy will attempt to route to thi
 
 In these cases, the request will be delivered to a different Machine that matches the remaining fields in your replay. Along with the other Fly.io-specific headers, a `fly-preferred-instance-unavailable` header will be set containing the ID of the instance that could not be reached.
 
+### Web Socket Considerations
+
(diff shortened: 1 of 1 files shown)


networking/dynamic-request-routing.html.markerb
```diff
@@ -241,6 +241,10 @@ If you replay with `prefer_instance` set, Fly Proxy will attempt to route to thi
 
 In these cases, the request will be delivered to a different Machine that matches the remaining fields in your replay. Along with the other Fly.io-specific headers, a `fly-preferred-instance-unavailable` header will be set containing the ID of the instance that could not be reached.
 
+### Web Socket Considerations
+
+It is worth noting that an application returning `fly-replay` headers should not negotiate a web socket upgrade itself.  Some frameworks automatically handle this process.  The application or instance to which the requests are forwarded should handle the upgrade instead.
```
kcmartin, line 246: The last sentence could be simplified a little, maybe:

```suggestion
It is worth noting that an application returning `fly-replay` headers should not negotiate a web socket upgrade itself.  Some frameworks automatically handle this process.  Instead, the application or instance receiving the requests should handle the upgrade.
```

kcmartin (approved):
LGTM!

PR #2249, opened 2025-09-30
Add new doc for Logs API / programmatic logs access options

kcmartin:
### Summary of changes

- Per request from @lubien 
- Add new doc to Logging section covering Logs API options (for programmatic access to logs)
- Update menus and overview pages in Monitoring/logging section

### Preview
Screenshot 2025-09-30 at 2 13 59 PM

### DIFF
--- a/monitoring/logs-api-options.html.md
+++ b/monitoring/logs-api-options.html.md
@@ -43,6 +43,8 @@ This endpoint isn’t officially documented, but it’s mostly stable: `flyctl` 
 
 Use this for quick fetches or simple polling scripts. If you hit rate limits or auth issues, check that your token has `read` access to the app.
 
+Importantly, this is the only option that gives you access to historical logs going back to the current retention window (about 15 days). The other options only start streaming from the moment they're connected.
+
 ### 2. Subscribe to logs via NATS (experimental)
 
 Logs are streamed through NATS under subjects like:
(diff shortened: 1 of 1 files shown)


monitoring/logs-api-options.html.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Options for fetching logs programmatically
+
+You can get logs programmatically via:
+
+- **Undocumented HTTP API**: What `fly logs` uses under the hood.
+- **NATS proxy**: Subscribe to real-time logs at the message level.
+- **Log shipper to external sink**: Ingest logs elsewhere, then query them via that system.
+
+Each has tradeoffs. If you want durable search, export logs. If you want low-latency push, use NATS. If you want to `curl` something now, use the HTTP API.
```
Roadmaster, line 25: I would say the main use case for the API is that it allows you fetch _historical_ logs up to the retention period (15 days or so). The other two options only start streaming logs from the moment you set them up, but they won't allow you to look at older logs.

monitoring/logs-api-options.html.md
```diff
@@ -0,0 +1,112 @@
...
+
+- You’ll need to handle reconnections: if the network drops or the proxy restarts, your subscriber needs to reconnect and resume without losing messages.
+- You’ll need to handle backpressure: if your app can’t process logs fast enough, messages might pile up and get dropped. Design your consumer to either keep up or fail gracefully.
+- If you want to dedupe across subscribers, use NATS queue groups.
+
+For more details on connecting to Fly’s NATS log stream (authentication, subject patterns, example clients), head over to the [Observability for User Apps](/docs/blueprints/observability-for-user-apps/?utm_source=chatgpt.com#streaming-fly-app-logs-to-your-end-users) guide.
+
+### 3. Log Shipper to External Sink
+
+If you want durable, queryable logs, export them. We maintain a small app called the [Fly Log Shipper](https://github.com/superfly/fly-log-shipper) that listens to NATS and forwards logs to sinks like Loki, Datadog, Elastic, Honeycomb, or even S3.
```
Roadmaster, line 78: We also have the Fly observability stack (https://github.com/superfly/fly-telemetry) - maybe we could ask around to see how much this is the way forward (tm) and recommend that instead of (or in addition to) the log shipper?

Roadmaster (approved):
two comments! +1 otherwise.

monitoring/logs-api-options.html.md
kcmartin: Thanks, I'll post internally and cover it in an update!

Roadmaster (approved):
LGTM!

PR #2270, opened 2025-10-16
Remove deprecated regions from  reference/regions

swilson-fly:
Removed all to-be-depreciated regions from the general region list.


swilson-fly:
Failing check seems to be it doesn't like `sao` in `sao paulo`

PR #2278, opened 2025-10-23
Update configuration.html.markerb

lubien:
### Summary of changes
- update the vm config example to match pricing table configs


lubien:
@kcmartin hmm good question, usually I expect one to go from shared-cpu-1x@256 to shared-cpu-1x@1024 so I dont have opinions, but Im up for 512MB wdyt?

lubien:
@kcmartin done!

kcmartin (approved):
LGTM!

PR #2285, opened 2025-10-28
Amend multi-container Machines docs

lesliebc:
### Summary of changes

Fixed some inconsistencies with the Machines API spec and updated "Deploying Multi-container Machines" section to clarify workflows

#### API spec corrections
Per [fly.ContainerConfig](https://docs.machines.dev/#model/flycontainerconfig):
- Fixed health check field name: `health_checks` -> `healthchecks` 
- Fixed so health check `grace_period`, `interval`, and `timeout` are integers e.g `"10s"` -> `10`
- Removed unsupported `services` field from container config

#### Other corrections
- Removed top-level `config` wrapper from `cli-config.json` examples
  - `machine-config` doesn't work unless `containers[]` is at the top level i.e `{ "containers": [...] }`
- Removed non-working `fly machine create` command example
  - the command shown in the docs returns `Error: requires at least 1 arg(s), only received 0` because it expects an `image` argument

#### Improvements
- Added "Using the Machines API" section
- Documented correct JSON config structure for API and CLI workflows

### DIFF
--- a/machines/guides-examples/multi-container-machines.html.markerb
+++ b/machines/guides-examples/multi-container-machines.html.markerb
@@ -166,7 +166,7 @@ The `config` object can also include other Machine fields like `guest` and `serv
 
 Refer to the [Machines API documentation](https://docs.machines.dev/#tag/machines) for complete configuration options.
 
-### Using flyctl
+### Using `fly machine run`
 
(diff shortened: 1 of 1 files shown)


kcmartin (approved):
Thank you for these changes and updates, LGTM!

PR #2321, opened 2025-12-17
Add documentation for app-scoped egress IPs

PeterCxy:
### Summary of changes

App-scoped egress IPs have been released for a few weeks under beta and seem to be working well, so document them here as the recommended solution for egress IPs.

Also changed the public networking page to make it clear that regular Anycast IPs are not used for egress and link to this page.

### DIFF
--- a/networking/egress-ips.html.md
+++ b/networking/egress-ips.html.md
@@ -39,7 +39,7 @@ fly ips allocate-egress --app <app-name> -r <region>
 This allocates a pair of static egress addresses, IPv4 and IPv6, for your app in a region.
 
 If your app has Machines in multiple regions, you must allocate at least 1 app-scoped static egress IP address __per region__.
-Only machines located in the same region can use static egress IPs allocated in the region.
+Machines can only use static egress IPs that were allocated in their own region.
 
 <div class="note icon">
 You can allocate multiple pairs of IPv4 and IPv6 static egress addresses in the same region. Machines will randomly choose a pair from all static egress IPs available in the region.


networking/egress-ips.html.md
```diff
@@ -9,10 +9,10 @@ date: 2025-10-02
 ## Overview
 
 - By default, outbound (egress) IPs from Fly Machines are **unstable** and may change.
-- You can allocate **static egress IPs** per machine (both IPv4 and IPv6) via `fly machine egress-ip`.
-- Static egress IPs come with trade-offs: cost, binding to machine lifecycle, and deployment quirks.
-- A common workaround is to front outbound traffic through a **proxy** app that _does_ have static egress IPs.
-- App-scoped egress IPs are in development and may simplify this in the future.
+- You can allocate **static egress IPs** for an app (both IPv4 and IPv6) via `fly ips allocate-egress`.
+- App-scoped Static egress IPs are per-region: you need one for every region where you have machines.
+- Static egress IPs come with trade-offs: cost, concurrent machines / connections.
```
kcmartin: Can you clarify this a bit more? Is the trade-off that they will hit limits for concurrent machines and connections? If that is the case maybe say something like: "Static egress IPs come with trade-offs: they cost more, and limit how many machines or connections you can run at once."

networking/egress-ips.html.md
```diff
@@ -26,8 +26,52 @@ Some external services—APIs, databases, payment providers—require allowlisti
 
 ---
 
+## Static Egress IPs (App-Scoped)
+
+App-scoped static egress IPs can be shared between multiple machines in a region belonging to the same app, and will not be deleted when machines are recreated. They are recommended over our legacy, machine-scoped static egress IPs.
```
kcmartin:
```suggestion
App-scoped static egress IPs can be shared between multiple machines in a region belonging to the same app, and will not be deleted when machines are recreated. They are recommended over our legacy machine-scoped static egress IPs.
```

kcmartin (changes requested):
Left a few suggested changes and an comment. Once those are addressed, we're good to go.

networking/egress-ips.html.md
```diff
@@ -26,8 +26,52 @@ Some external services—APIs, databases, payment providers—require allowlisti
...
+### Allocate an App-scoped Static Egress IP
+
+```bash
+fly ips allocate-egress --app <app-name> -r <region>
+```
+
+This allocates a pair of static egress addresses, IPv4 and IPv6, for your app in a region.
+
+If your app has Machines in multiple regions, you must allocate at least 1 app-scoped static egress IP address __per region__.
+Only machines located in the same region can use static egress IPs allocated in the region.
```
kcmartin:
```suggestion
Machines can only use static egress IPs that were allocated in their own region.
```

PR #2326, opened 2025-12-18
Revised edits for the new egress-ip (app scoped) feature

Lucais11:
### Summary of changes

1. Fixed typo
- "availble" → "available" in Overview section

2. Added beta status
- Section title updated to include "(Beta)" 

Why- The live doc treated app-scoped egress IPs as GA, but they're in beta. This should help set the proper expectations about stability and potential behavior changes that customers might experience.

3. Clarified release behavior
- Added explicit statement that app-scoped egress IPs are only released when you run `fly ips release-egress` and persist across Machine destruction.

Why- The live doc implied persistence but didn't state release behavior, which can cause billing confusion and leaked IPs.

4. Updated billing section
- Added beta pricing information- free during beta, expected to bill at $3.60/mo after beta.

Why: The live doc showed final pricing, but the feature is in beta and currently free. This prevents billing any surprises.

5. Standardized and improved Caveats section
- Connection limit standardized to 1024 concurrent connections (was 1000). This was written here as well originally- https://community.fly.io/t/beta-feature-try-out-app-scoped-egress-ips/26536.
- Changed "external IP address" to "destination IP address" for accuracy
- Fixed indentation of connection limit note
- Improved delay explanation, so now there is context about why delays occur
- Added a note about flyctl warnings when limits are approached, so users are aware

6. Added new section: "Interaction with Machine-Scoped Egress IPs"
- Explains that app-scoped and machine-scoped IPs shouldn't be used together
- States that machine-scoped IPs take precedence over app-scoped IPs
- Includes migration guidance with a warning callout

Why: During migration, machines can silently use machine-scoped IPs instead of app-scoped ones, causing unexpected routing. Want to make sure users are aware of this.

### DIFF
--- a/networking/egress-ips.html.md
+++ b/networking/egress-ips.html.md
@@ -62,9 +62,14 @@ Each app-scoped IPv4 static egress address costs $3.60/mo, billed hourly. IPv6 a
 
 - Each static egress IP can support up to 64 Machines. If you need more than 64 Machines in one region, you will need to allocate multiple static egress IPs.
 - When using app-scoped static egress IPs, a Machine can make up to 1024 concurrent connections to _each_ destination IP address. There is no limit on the _total_ number of concurrent connections.
-- We do not expect this to be a concern for most apps. However, feel free to talk to us if this limits your use case!
+
+<div class="note icon">
+We do not expect this to be a concern for most apps. However, feel free to talk to us if this limits your use case!
+</div>
 - When you have multiple static egress IPs assigned in one region, there is currently no way to specify exactly which IP each machine will use.
-- There may be a short delay between allocating an egress IP (or creating a Machine) and the egress IP becoming usable. This is inherent to how the system applies IPs to new Machines. This delay may be more noticeable with more Machines or during bluegreen deployments. Allocating multiple pairs of static egress IPs alleviates the issue.
+- There may be delays when egress IPs are applied to Machines:
+- Right after allocating a new egress IP, it will be applied to all existing Machines in the region after a short delay. Allocating multiple pairs of static egress IPs will not help in this case.
+- When creating a new Machine in an app that already has an egress IP assigned, there may be a delay before the Machine can use the egress IP. This delay may be more noticeable with more Machines or during bluegreen deployments. Allocating multiple pairs of static egress IPs can help alleviate this issue.
 - `flyctl` surfaces warnings when these limits are approached during Machine creation, deployments, and IP management.
 
 ### Interaction with Machine-Scoped Egress IPs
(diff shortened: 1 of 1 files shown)


kcmartin (approved):
LGTM

networking/egress-ips.html.md
```diff
@@ -52,17 +52,30 @@ fly ips list
...
+Each app-scoped IPv4 static egress address costs $3.60/mo, billed hourly. IPv6 addresses are currently free, but must be allocated along with an IPv4. During the beta period, app-scoped egress IPs are free. After the beta, app-scoped IPv4 egress IPs are expected to be billed at the same rate as machine-scoped egress IPs ($3.60/mo, billed hourly).
 
 ### Caveats
 
 - Each static egress IP can support up to 64 Machines. If you need more than 64 Machines in one region, you will need to allocate multiple static egress IPs.
-- When using App-scoped static egress IPs, a Machine can make up to 1000 connections to _each_ external IP address. There is no limit on the _total_ number of concurrent connections.
-  - We do not expect this to be a concern for most apps. However, feel free to talk to us if this limits your use case!
+- When using app-scoped static egress IPs, a Machine can make up to 1024 concurrent connections to _each_ destination IP address. There is no limit on the _total_ number of concurrent connections.
+- We do not expect this to be a concern for most apps. However, feel free to talk to us if this limits your use case!
```
PeterCxy: This should have one more level of indentation?

networking/egress-ips.html.md
```diff
@@ -52,17 +52,30 @@ fly ips list
...
 - When you have multiple static egress IPs assigned in one region, there is currently no way to specify exactly which IP each machine will use.
-- When new machines are created, there might be a brief window when an app-scoped egress IP is not applied to the machine. This may happen more often with more machines or during bluegreen deployment. Allocating multiple pairs of static egress IPs alleviates the issue.
+- There may be a short delay between allocating an egress IP (or creating a Machine) and the egress IP becoming usable. This is inherent to how the system applies IPs to new Machines. This delay may be more noticeable with more Machines or during bluegreen deployments. Allocating multiple pairs of static egress IPs alleviates the issue.
```
PeterCxy: I don't think this description is entirely accurate. It's essentially talking about two possible scenarios:

- Right after allocating a new egress IP, in that case the egress IP will be applied to all existing machines after a short delay. Allocating multiple pairs will probably _not_ help in this case (since allocating them might be slower :smile: ).
- The app already has an egress IP assigned, and a new machine is created. In that case there might be a delay before the machine gets to use the egress IP, and only in this case might allocating multiple pairs help.

PR #2328, opened 2025-12-24
add rama as rust framework to fly.io docs

GlenDC:
### Summary of changes

Adds documentation on how to deploy rama apps to fly.io.
Long overdue given how we ourselves build also all our public
services on top of fly.io as well as some of our private ones.

### DIFF
--- a/rust/frameworks/rama.html.markerb
+++ b/rust/frameworks/rama.html.markerb
@@ -1,0 +1,73 @@
+---
+title: "Run a Rama App"
+layout: framework_docs
+objective: Rama is a modular service framework to move and transform network packets.
+redirect_from:
+ - /docs/languages-and-frameworks/rama/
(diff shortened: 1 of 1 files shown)


kcmartin:
Hi @GlenDC, sorry for the long delay here, and thank you for putting this together. rama is a neat framework and it's great to see it running so well on Fly.

We've decided to hold off on adding a dedicated rama guide for now. We try to keep our framework guides limited to ones we're confident we can maintain over time, and rama's fast-moving API makes that a tougher commitment on our end (the snippet here has already drifted from the current 0.3.0 release since the PR was opened).

We'd still love for rama users to find this though. Our [Community forum](https://community.fly.io/) is a great home for a deploy walkthrough, and we're always happy to point people there. Thanks again for building on Fly!

PR #2371, opened 2026-03-25
docs: add client configuration guides for MPG

jphenow:
### Summary of changes

This adds comprehensive documentation for configuring Managed Postgres client connections, addressing a critical gap where users lacked guidance on preventing dropped connections during proxy maintenance.

The core problem: Fly's edge proxy restarts periodically (typically during deployments), and connections held longer than the proxy's shutdown timeout are forcibly closed. Without proper configuration, most applications encounter `ECONNRESET` or `tcp recv (idle): closed` errors during these restarts. Connection pool configuration is the fix, but it was poorly documented across the MPG docs.

This change adds two guides. "Connect Your Client" is a quick-start checklist covering the essential settings: max connection lifetime (600s), idle timeout (300s), pool size recommendations, and prepared statement handling. "Client-Side Connection Configuration" is a comprehensive reference with language-specific examples for Node.js, Python, Go, Ruby, and Elixir, detailed explanations of how PgBouncer modes affect client behavior, connection limit tables per plan tier, and a full troubleshooting section addressing common errors.

The navigation has been restructured to surface both guides prominently, and existing guides (cluster configuration, Phoenix) now cross-link to the new documentation. This makes the recommended connection settings discoverable at the point where users need them most — when connecting their applications to MPG.

### Notes

* [ ] TODO: some numbers like pool counts are incorrect
* [ ] reread through this, this is a quick dump of a smattering of conversations so it needs some more verification


mpg/client-configuration.html.md
```diff
@@ -0,0 +1,307 @@
...
+
+The proxy's shutdown timeout is **10 minutes**. Any connection that remains open after that is terminated. If your application holds connections for longer than this — which is the default behavior of most connection pools — you might run into errors like `tcp recv (idle): closed` or `ECONNRESET` during proxy deployments.
+
+The fix is straightforward: configure your connection pool to **proactively recycle connections** on a shorter interval than the proxy's timeout.
+
+## Recommended settings
+
+| Setting | Recommended value | Why |
+|---------|-------------------|-----|
+| Max connection lifetime | **600s** (10 min) | Connections recycle before the proxy's 15-min shutdown timeout |
```
PeterCxy: > Connections recycle before the proxy's 15-min shutdown timeout

Let's not say this publicly haha, that's not a guarantee. What about

> Recycle connections to avoid disconnects during `fly-proxy` restart

we can link to fly-proxy's documentation page here too

jphenow: Oh yea, I rewrote one but missed this. Need to reread through it again in the morning

Roadmaster:
Hi, Django is one of our best-supported Python frameworks. Is it possible to please add a section on Django? CONN_MAX_AGE=600 is the thing to add to use a pool of persistent connections _and_ recycle them after 10 minutes. The default is 0 which means a new connection for each request and tearing it down when the request context is closed.

PR #2378, opened 2026-04-09
fix unmanaged psql image update explanation

kubosuke:
### Summary of changes

image update happens only when digest has changed within the same image:tag 

https://github.com/superfly/flyctl/blob/44d56a688d1a979601017402965a6367039b2b72/internal/command/image/update_machines.go#L278-L301

https://fly.io/docs/postgres/managing/upgrades/


kubosuke:
https://github.com/superfly/flyctl/issues/4812#event-24779751296

PR #2380, opened 2026-04-10
add docker compose section to multicontainer guide

kcmartin:
### Summary of changes
Add a self-contained section to add to the multi-container Machines guide. Covers config, routing, a practical nginx + Redis example, three clear limitations.

### Preview
Screenshot 2026-04-10 at 4 16 53 PM

### DIFF
--- a/machines/guides-examples/multi-container-machines.html.markerb
+++ b/machines/guides-examples/multi-container-machines.html.markerb
@@ -229,6 +229,10 @@ fly deploy
 
 The `web` container handles inbound traffic on port 80, and the `redis` container runs alongside it as a sidecar. Your application code can reach Redis at `localhost:6379` because both containers share the same network namespace inside the Machine.
 
+<div class="warning icon">
+**Warning:** Fly ignores `volumes:` declarations in your Compose file and prints `Warning: Could not read volume file...` at deploy time. The deploy still succeeds, but any data your containers write is stored on an ephemeral overlay that's wiped on restart or redeploy. For persistent storage, attach a [Fly Volume](/docs/volumes/) and mount it via `[mounts]` in `fly.toml`.
+</div>
+
 #### Limitations
 
 - **Exactly one buildable service.** `fly deploy` requires exactly one service in the Compose file to specify `build`. All other services must use pre-built images.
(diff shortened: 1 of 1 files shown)


Roadmaster (approved):
LGTM! I tested it just to see it in action and confirm these instructions work - they do, out of the box.

Thanks!

Actually - EDIT: The compose file I used [here](https://raw.githubusercontent.com/paperless-ngx/paperless-ngx/refs/heads/main/docker/compose/docker-compose.sqlite.yml) specifies volumes, but flyctl will happily ignore volume setup it seems, and this _will_ lead to data loss (imagine someone deploys paperless-ngx using that, writes a bunch of data to the rootfs and then restarts the machine).

We don't need to hold this doc update until we figure this out, but it _would_ perhaps be good to point out somewhere that some compose configurations are recognized  and handled by flyctl and some are happily ignored.

FWIW flyctl _did_ say something about volumes at deploy time, but it's not super clear that this means "i'm not going to create or mount these volumes for you" - in principle it could not, because a Fly machine can only have one attached volume and this configuration wants 5 volumes.

```
$ fly deploy --ha=false
==> Verifying app config
Warning: Could not read volume file redisdata: open redisdata: no such file or directory
Warning: Could not read volume file data: open data: no such file or directory
Warning: Could not read volume file media: open media: no such file or directory
Warning: Could not read volume file export: open export: no such file or directory
Warning: Could not read volume file consume: open consume: no such file or directory
```

In summary, I think just a quick warning that if your compose file specifies volumes it _will_ need adapting by hand might be worth it.

kcmartin:
@Roadmaster 

> Actually - EDIT: The compose file I used [here](https://raw.githubusercontent.com/paperless-ngx/paperless-ngx/refs/heads/main/docker/compose/docker-compose.sqlite.yml) specifies volumes, but flyctl will happily ignore volume setup it seems, and this _will_ lead to data loss (imagine someone deploys paperless-ngx using that, writes a bunch of data to the rootfs and then restarts the machine).

> In summary, I think just a quick warning that if your compose file specifies volumes it _will_ need adapting by hand might be worth it.

Thanks for finding this. I tested and confirmed the following: 

I reproduced the behavior with a volumes: declaration in the Compose file and confirmed:

  - fly deploy prints Warning: Could not read volume file... multiple times but succeeds
  anyway.
  - Data written to the path intended for the volume lives on an ephemeral overlay — it's
  wiped on both fly machine restart and subsequent fly deploy runs.

 I've added a warning callout just above the Limitations section pointing this out and directing readers to Fly Volumes + `[mounts]` for persistent storage.

Roadmaster:
LGTM thanks! ship it!

PR #2387, opened 2026-04-27
Add guidance on splitting process groups into separate apps

kcmartin:
## Summary
Expands the "Maybe you want multiple apps, though" section in `app-guides/multiple-processes.html.md` with practical guidance on when and how to split process groups into separate Fly apps:

- Calls out the metrics-based autoscaler as a concrete case where separate apps are required (whole-app scaling)
- Documents the shared-image workflow: `fly deploy --build-only --push` to build once, then `fly deploy --image registry.fly.io/app:deployment-xxxx` to deploy to each app
- Notes the secrets-duplication caveat

Content sourced from the internal solutions architecture guide. Inter-app `.internal` networking was already covered in the preceding paragraph, so no duplication.


injoongy (approved):
LGTM 🚀

PR #2394, opened 2026-04-29
Rewrite Flycast blueprint to drop GPU/Ollama example

kcmartin:
## Summary

GPU Machines are being deprecated on 2026-08-01, and the Flycast blueprint at `blueprints/private-applications-flycast.html.md` was built almost entirely around an Ollama-on-GPU walkthrough. The Flycast concepts themselves are GPU-independent, so this rewrites the guide rather than removing it.

- Replaces the Ollama example with a generic private HTTP service using the public `nginxdemos/hello` image, so no external example repo is needed
- Swaps the multi-step `fly launch --no-deploy` + `fly ips allocate-v6 --private` + `fly deploy` flow for the modern `fly launch --image ... --flycast`
- Removes the YouTube embed (GPU/Ollama walkthrough), the "Running Llama 3" subsection, and the GPU-related related-reading links
- Replaces "Fly Postgres" with "Managed Postgres" in the security note (the currently supported offering)
- Refreshes related reading to point at the autostart blueprint, the Flycast reference, and private networking
- Fixes a separately broken link (`/docs/reference/services/#ip-addresses` → `/docs/networking/services/#anycast-ip-addresses`); the old path no longer resolves
- Adds `walkthrough` to `styles/config/vocabularies/fly-terms/accept.txt` so Vale stops flagging it as a typo (it's already used elsewhere in the docs)

## Walkthrough gotchas baked into the recipe

- `--internal-port 80` on `fly launch`: `nginxdemos/hello` listens on 80, but Fly Launch defaults to 8080
- `force_https = false` in `fly.toml`: Flycast addresses don't have public TLS certs, so the default 301-to-HTTPS breaks the demo's plain HTTP requests
- `--vm-memory 1024` on the shell Machine: the 256MB default OOM-kills during `apt install -y curl iputils-ping dnsutils`

### DIFF
--- a/blueprints/private-applications-flycast.html.md
+++ b/blueprints/private-applications-flycast.html.md
@@ -6,10 +6,8 @@ nav: guides
 author: xe
 categories:
   - networking
-date: 2024-06-17
+date: 2026-04-29
 ---
-
-<center><iframe width="600" height="315" src="https://www.youtube-nocookie.com/embed/PR0rzkwKwFA" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe></center><br />
 
 ## Overview
 
@@ -17,7 +15,7 @@ A lot of the time your applications are made to be public and shared with the wo
 
 Sometimes you need a middle ground between fully public apps and fully private apps. [Flycast](/docs/networking/flycast/) addresses are private but global IPv6 addresses inside your private network that go through the Fly Proxy, so you get all of the load management and Machine waking powers that you get for free with public apps.
 
-This blueprint covers what Flycast is, when and why you’d want to use it, and shows you how to create an instance of Ollama that you can connect to over Flycast.
+This blueprint covers what Flycast is, when and why you’d want to use it, and shows you how to deploy a small private HTTP service that you can connect to over Flycast.
 
 ## What is Flycast?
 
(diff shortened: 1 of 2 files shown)


blueprints/private-applications-flycast.html.md
```diff
@@ -56,41 +50,47 @@ To get started, you need to do the following:
...
 ```
-fly launch --from https://github.com/fly-apps/ollama-demo --no-deploy
+fly launch --image nginxdemos/hello --flycast --internal-port 80 --no-deploy
 ```
 
-This command creates a new fly app from the [`ollama-demo` template](https://github.com/fly-apps/ollama-demo+external) and tells the flyctl command to not deploy it after you create the app. If we don’t do this, then the platform will create public IPv4 and IPv6 addresses, which will make this a public app. The name you choose when you create your app will be used to connect to your app over Flycast.
+The name you choose for your app will be the hostname you use to reach it over Flycast (`<appname>.flycast`). When `fly launch` asks if you want to tweak the settings, you can accept the defaults.
 
-Next, allocate a Flycast address for your app with the `fly ips allocate-v6` command:
+The `--internal-port 80` flag tells Fly Launch that our app listens on port 80 (the default for the `nginxdemos/hello` image). Fly Launch defaults to port 8080, so without this flag the Fly Proxy wouldn't be able to reach the app.
```
Roadmaster: I don't know if adding a clarifying "without this flag or manually tweaking the config file proxy won't be able to reach the app" makes sense here or muddies the waters.

Roadmaster: Since you're going to edit the toml file in the next step anyway, perhaps it's worth mentioning that the port could be changed either way.

Roadmaster (approved):
Awesome! Thanks! One small nitpick.

blueprints/private-applications-flycast.html.md
kcmartin: Thanks, I agree that's worth a clarification, and I cleaned it up.

PR #2399, opened 2026-05-07
Document one-app-per-user scale limit for autostop

kcmartin:
### Summary of changes

Add a footgun to the [per-user dev environments](https://fly.io/docs/blueprints/per-user-dev-environments/) blueprint covering what happens when you put thousands of user Machines into a single app: the Fly Proxy's autostop loop is rate-limited (one Machine per region per pass, every few minutes) and can't keep up at that scale, so most idle Machines stay running. Also notes the shared-secrets / flat-network consequence and points readers to stop-when-idle as the alternative.

Cross-links from the [autostop config guide](https://fly.io/docs/launch/autostop-autostart/) and the [Fly Proxy autostop reference](https://fly.io/docs/reference/fly-proxy-autostop-autostart/) back to the dynamic-routing pattern, so readers landing on the autostop docs while running this shape are pointed to a workable approach.

### Notes

Came out of a support thread where a customer running thousands of Machines in one app discovered the rate-limited stop loop the hard way. The behavior is already documented in the proxy reference; this just makes the implication explicit at the per-user-dev page and adds the reverse pointers.


Roadmaster (approved):
LGTM thanks!

PR #2407, opened 2026-05-24
Update logging docs for VictoriaLogs migration and app logs API corrections

lesliebc:
### Summary of changes

Updates logging docs to reflect migration from Quickwit to VictoriaLogs. References to Quickwit are replaced with VictoriaLogs, and our log retention period is updated to 7 days.

Also makes a few changes to the app logs endpoint doc:
- Changes API response format from newline-delimited JSON stream to JSON:API document
- Replaces `start_time` with `next_token` in supported query params
- Documents endpoint's default behaviour

### DIFF
--- a/monitoring/logs-api-options.html.md
+++ b/monitoring/logs-api-options.html.md
@@ -41,7 +41,7 @@ The response is a JSON:API document with a `data` array of log entries and a `me
 
 - `region`: three-letter region code. Returns only logs from this region
 - `instance`: a Machine ID. Returns only logs from this instance
-- `next_token`: a nanosecond Unix timestamp. Returns logs after this time. Each response includes a `meta.next_token` — pass it in your next request to page forward
+- `next_token`: a nanosecond Unix timestamp (e.g `1779235200000000000`). Returns logs after this time. Each response includes a `meta.next_token` — pass it in your next request to page forward
 
 This endpoint isn’t officially supported for external use, but it’s mostly stable: `flyctl` depends on it.
 
--- a/monitoring/logs-api-options.html.md
+++ b/monitoring/logs-api-options.html.md
@@ -37,13 +37,13 @@ GET /api/v1/apps/:app_name/logs
 Authorization: <your-token>
 ```
 
-You’ll get a stream of newline-delimited JSON log lines. You can pass query params like:
+The response is a JSON:API document with a `data` array of log entries and a `meta.next_token` cursor for paging. By default, the most recent 100 logs from the last 24 hours are returned. You can pass these query params to filter logs or set a starting timestamp:
 
-- `region=cdg` (filter by region)
-- `instance=<id>` (filter by instance — sometimes flaky)
-- `start_time=2023-08-01T00:00:00Z` (to backfill)
+- `region`: three-letter region code. Returns only logs from this region
+- `instance`: a Machine ID. Returns only logs from this instance
+- `next_token`: a nanosecond Unix timestamp. Returns logs after this time. Each response includes a `meta.next_token` — pass it in your next request to page forward
 
-This endpoint isn’t officially documented, but it’s mostly stable: `flyctl` depends on it. That said, filters don’t always work as expected.
+This endpoint isn’t officially supported for external use, but it’s mostly stable: `flyctl` depends on it.
 
 Use this for quick fetches or simple polling scripts. If you hit rate limits or auth issues, check that your token has `read` access to the app.
 
@@ -78,7 +78,7 @@ This gives you structured JSON log messages in real time.
 - If you want to dedupe across subscribers, use NATS queue groups.
 - NATS only streams logs from starting from the moment you connect. You won’t get any history unless you’ve been subscribed the whole time.
 
-For more details on connecting to Fly’s NATS log stream (authentication, subject patterns, example clients), head over to the [Observability for User Apps](/docs/blueprints/observability-for-user-apps/?utm_source=chatgpt.com#streaming-fly-app-logs-to-your-end-users) guide.
+For more details on connecting to Fly’s NATS log stream (authentication, subject patterns, example clients), head over to the [Observability for User Apps](/docs/blueprints/observability-for-user-apps/#streaming-fly-app-logs-to-your-end-users) guide.
 
 ### 3. Log Shipper to External Sink
 
--- a/monitoring/logs-api-options.html.md
+++ b/monitoring/logs-api-options.html.md
@@ -12,7 +12,7 @@ date: 2025-10-01
 
 ## Overview
 
-Fly.io apps run on Firecracker microVMs we call Machines. Each Machine captures `stdout`/`stderr`, ships those logs over NATS, and stores them for a period of time in a Quickwit-backed search index. Most users consume logs via `fly logs` or by setting up log shipping to an external sink.
+Fly.io apps run on Firecracker microVMs we call Machines. Each Machine captures `stdout`/`stderr`, ships those logs over NATS, and stores them for a period of time in a VictoriaLogs-backed search index. Most users consume logs via `fly logs` or by setting up log shipping to an external sink.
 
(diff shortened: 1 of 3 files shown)


monitoring/logs-api-options.html.md
```diff
@@ -37,17 +37,17 @@ GET /api/v1/apps/:app_name/logs
...
 
-- `region=cdg` (filter by region)
-- `instance=<id>` (filter by instance — sometimes flaky)
-- `start_time=2023-08-01T00:00:00Z` (to backfill)
+- `region`: three-letter region code. Returns only logs from this region
+- `instance`: a Machine ID. Returns only logs from this instance
+- `next_token`: a nanosecond Unix timestamp. Returns logs after this time. Each response includes a `meta.next_token` — pass it in your next request to page forward
 
-This endpoint isn’t officially documented, but it’s mostly stable: `flyctl` depends on it. That said, filters don’t always work as expected.
+This endpoint isn’t officially supported for external use, but it’s mostly stable: `flyctl` depends on it.
```
kcmartin, line 47: The old text included "That said, filters don't always work as expected." The new version removes it without replacement. Is this because the VictoriaLogs migration actually fixed filter reliability, or just because the bullet was being rewritten? If filters are still occasionally flaky maybe worth keeping a one-line caveat.

monitoring/search-logs.html.markerb
```diff
@@ -6,20 +6,14 @@ redirect_from: /docs/metrics-and-logs/search-logs/
 ---
 
 <div class="warning icon">
-Searchable application logs is a beta feature. Learn more and post your suggestions or issues in the community thread: [Searchable application logs in Grafana](https://community.fly.io/t/searchable-application-logs-in-grafana/18878+external)
+Searchable application logs is a beta feature. During the beta period, log search is free.
```
kcmartin:
```suggestion
Searchable application logs are a beta feature. During the beta period, log search is free.
```

monitoring/logs-api-options.html.md
```diff
@@ -37,17 +37,17 @@ GET /api/v1/apps/:app_name/logs
...
 
-You’ll get a stream of newline-delimited JSON log lines. You can pass query params like:
+The response is a JSON:API document with a `data` array of log entries and a `meta.next_token` cursor for paging. By default, the most recent 100 logs from the last 24 hours are returned. You can pass these query params to filter logs or set a starting timestamp:
 
-- `region=cdg` (filter by region)
-- `instance=<id>` (filter by instance — sometimes flaky)
-- `start_time=2023-08-01T00:00:00Z` (to backfill)
+- `region`: three-letter region code. Returns only logs from this region
+- `instance`: a Machine ID. Returns only logs from this instance
+- `next_token`: a nanosecond Unix timestamp. Returns logs after this time. Each response includes a `meta.next_token` — pass it in your next request to page forward
```
kcmartin: `next_token` is described as both "a starting timestamp" (user-suppliable) and "pass it in your next request to page forward" (opaque cursor). Can you confirm which is the intended use, and update as needed?
- If users can construct one themselves to backfill from an arbitrary moment: add an example value to the bullet (the old doc had `start_time=2023-08-01T00:00:00Z` what's the nanosecond equivalent look like?).
- If it's only an opaque cursor returned by the previous response: drop "or set a starting timestamp" from the intro sentence on line 40, and rewrite the bullet as something like "an opaque paging cursor returned in `meta.next_token` of the previous response."

Either way, the rename from `start_time` (ISO-8601) to `next_token` (nanosecond Unix) is a breaking change for anyone using the old param. A one-line note calling that out would save existing users some debugging.

monitoring/search-logs.html.markerb
```diff
@@ -6,20 +6,14 @@ redirect_from: /docs/metrics-and-logs/search-logs/
...
-You can access searchable logs in Grafana:
+To search logs from your Fly Apps:
 
-- Click **Log Search** in your Fly.io app dashboard, or
-- Go directly to your [Grafana metrics](https://fly-metrics.net/d/fly-logs/fly-logs+external)
+- Click **Search logs in Grafana** in your Fly.io app dashboard, or
+- Open your [Grafana dashboard](https://fly-metrics.net/d/fly-logs/fly-logs) directly
 
-<div class="important icon">
-**Important:** If you’re already signed in to Grafana, then you'll need to log out and log back in.
```
kcmartin, line 20: Confirming, do we no longer need users to log out and log back in?

kcmartin (changes requested):
Left a few comments and a suggestion

monitoring/logs-api-options.html.md
lesliebc: I removed it because the filters _aren't_ flaky in my experience (they're what flyctl uses for `fly logs -i` and `fly logs -r`. I'll put the caveat back in if you've experienced otherwise and think it's better to keep that in.

monitoring/search-logs.html.markerb
lesliebc: Logging out/in was needed to refresh the Grafana UI and expose log search when we first launched this with Quickwit in 2024. I removed the call out because I think, by now, most users will have reauthed and can access log search today. Can re-add if you think that's better!

kcmartin: Let's keep your removal of the call-out, as that seems like enough time has passed for users to have completed that action.

monitoring/logs-api-options.html.md
kcmartin: I don't have experience to add to that. We can always add more details, if we hear evidence to the contrary. I think we're fine leaving it at 'isn't officially supported for external use".

monitoring/logs-api-options.html.md
lesliebc: `next_token` can be used both ways that you described. I think it's primarily a cursor, but you can also approximate backfilling by passing in your own timestamp as `next_token`. Could the usage be clearer? I also added an example for that param like you suggested. Thanks!

About the previously documented `start_time`, I don't know that this was working before. I don't see that param in the bubblegum or web code, but I could've missed it. `next_token` has been in the logging code for a long time - I think this was always the supported way to backfill and page through logs, before and since migrating to VictoriaLogs.

Edit: I added a bullet after `next_token` stating that it replaces `start_time` to avoid confusion. Let me know if anything should be tweaked. Happy to update.

monitoring/logs-api-options.html.md
```diff
@@ -37,17 +37,18 @@ GET /api/v1/apps/:app_name/logs
...
 
-You’ll get a stream of newline-delimited JSON log lines. You can pass query params like:
+The response is a JSON:API document with a `data` array of log entries and a `meta.next_token` cursor for paging. By default, the most recent 100 logs from the last 24 hours are returned. You can pass these query params to filter logs or set a starting timestamp:
 
-- `region=cdg` (filter by region)
-- `instance=<id>` (filter by instance — sometimes flaky)
-- `start_time=2023-08-01T00:00:00Z` (to backfill)
+- `region`: three-letter region code. Returns only logs from this region
+- `instance`: a Machine ID. Returns only logs from this instance
+- `next_token`: a nanosecond Unix timestamp (e.g `1779235200000000000`). Returns logs after this time. Each response includes a `meta.next_token` — pass it in your next request to page forward
```
kcmartin: Could we make the dual purpose of next_token (cursor + timestamp backfill) explicit in the bullet? Something like:


```suggestion
- `next_token`: a nanosecond Unix timestamp (e.g. `1779235200000000000`) used either as a paging cursor or to backfill from a specific moment. Returns logs after this time. Each response includes a `meta.next_token` that you can pass in your next request to page forward.
```

kcmartin (changes requested):
Thanks for the updates. There is one minor suggestion for clarity + fixing a typo, otherwise LGTM!

kcmartin (approved):
LGTM! Thank you!

monitoring/logs-api-options.html.md
lesliebc: Good call, thanks for your review and all of your suggestions!

PR #2412, opened 2026-06-10
Fix Extensions pricing: Tigris data transfer is billed

Lucais11:
PR #2196 removed the statements exempting Tigris Object Storage from data-transfer billing but left a "Bandwidth to Tigris Object Storage" bullet in the "not billed separately for" list, contradicting the billed sentence below it. Remove that leftover bullet and name Tigris explicitly in the billed sentence, so the page consistently states that data transfer to third-party services, including Tigris, is billed.


Lucais11:
Vale linter error is irrelevant here 😄

kcmartin (approved):
LGTM

PR #2416, opened 2026-06-15
Add long-running tasks and machine lifecycle blueprint

kcmartin:
## Summary

New blueprint covering the interaction between `auto_stop_machines` and long-running in-process work. Documents how the Fly proxy decides to stop a machine, why background tasks are invisible to that decision, and two patterns to keep work from getting killed:

- **Pattern A:** disable autostop and manage shutdown in the app via `SIGTERM`/`kill_timeout`.
- **Pattern B:** split web and worker into separate process groups so autostop only applies to the web tier.

Also covers `kill_signal`/`kill_timeout` semantics under autostop and other stop pathways (deploys, `fly machine stop`, host migrations), and a Common Problems section that addresses things like self-pings (they don't work) and workers not stopping on deploy.

## Placement

- New file: `blueprints/long-running-tasks.html.md`
- Index entry: added to **Background Jobs & Automation** in `blueprints/index.html.md` with a `NEW!!` tag, next to the work-queues, task-scheduling, and supercronic blueprints.
- Sidebar: added to the matching **Background Jobs & Automation** group in `partials/_guides_nav.html.erb`.

## Empirical backing

Every technical claim in the draft is backed by a live deployment test on Fly. Specifically:

- Pattern A confirmed working: `auto_stop_machines = "off"` keeps machines up; `SIGTERM`/`kill_timeout` graceful drain fires on manual stop with the full drain window observed.
- Pattern B confirmed working: with split process groups, the proxy stops the `web` tier while the `worker` tier stays untouched.
- Self-ping (an earlier "Pattern C" candidate) tested and **does not** prevent autostop, even with successful HTTP requests every 60s to the machine's own `<app>.fly.dev` hostname. The proxy stops the machine within 5 to 10 minutes regardless. This is captured in the Common Problems section as "Why doesn't a self-ping keep my machine alive?"
- `kill_timeout` confirmed to be honored under the autostop pathway (full drain window before `SIGKILL`).

### DIFF
--- a/blueprints/index.html.md
+++ b/blueprints/index.html.md
@@ -62,11 +62,11 @@ Make it fast. Make it reliable. Monitor what's happening.
 
 How to run periodic tasks, long-running jobs, infrastructure automation, and the things that run when you’re asleep.
 
+- [Long-running tasks and machine lifecycle](/docs/blueprints/long-running-tasks/) NEW!!
 - [Building Infrastructure Automation without Terraform](/docs/blueprints/infra-automation-without-terraform/)
 - [Deferring long-running tasks to a distributed work queue](/docs/blueprints/work-queues/)
 - [Task scheduling guide with Cron Manager and friends](/docs/blueprints/task-scheduling/)
 - [Crontab with Supercronic](/docs/blueprints/supercronic/)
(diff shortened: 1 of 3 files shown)


injoongy (approved):
LGTM 🚀

PR #2417, opened 2026-06-19
Clarify max amount and hold time for preauthorization charges

Roadmaster:
A user pointed out the documentation shows amounts and hold times that don't match what the UX says (ui-ex) nor what happens in reality. This updates the documentation so it matches actual behavior and/or other information we give users.

### DIFF
--- a/about/billing.html.markerb
+++ b/about/billing.html.markerb
@@ -119,7 +119,7 @@ We require an active, valid credit card on file for most Fly.io accounts to do t
 
 We may send a pre-authorization request to the issuing bank to verify a bank will authorize charges. Some credit card companies present these as real charges, which may surprise you.
 
-A pre-auth is really just a hold on a credit card; it is not a real charge. We’ll typically pre-authorize a small amount (usually less than US $10) after signup, then cancel the authorization immediately. Banks may show the pre-auth for up to 10 working days, even though we have not collected any money.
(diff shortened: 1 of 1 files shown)


about/billing.html.markerb
```diff
@@ -119,7 +119,7 @@ We require an active, valid credit card on file for most Fly.io accounts to do t
 
 We may send a pre-authorization request to the issuing bank to verify a bank will authorize charges. Some credit card companies present these as real charges, which may surprise you.
 
-A pre-auth is really just a hold on a credit card; it is not a real charge. We’ll typically pre-authorize a small amount (usually less than $5) after signup, then cancel the authorization immediately. Banks may show the pre-auth for up to 7 days, even though we have not collected any money.
+A pre-auth is really just a hold on a credit card; it is not a real charge. We’ll typically pre-authorize a small amount (usually less than US $10) after signup, then cancel the authorization immediately. Banks may show the pre-auth for up to 10 working days, even though we have not collected any money.
```
kcmartin:
```suggestion
A pre-auth is really just a hold on a credit card; it is not a real charge. We’ll typically pre-authorize a small amount (usually less than US $10) after signup, then cancel the authorization immediately. Banks may show the pre-auth for up to 10 business days, even though we have not collected any money.
```

kcmartin (approved):
Small change suggested, I believe "business days" is more common.

Otherwise, LGTM

PR #2420, opened 2026-07-01
Secure log shipper docs: no public services or IPs

kcmartin:
The Log Shipper guide told users to edit `fly.toml` and keep a `[[services]]` section on `internal_port = 8686`. That port serves Vector's GraphQL API (including an unauthenticated `/playground`), so following the docs exposed it publicly. This was [reported on the community forum](https://community.fly.io/t/fly-log-shipper-publishes-graphql-publicly/14836), where Fly staff recommended deploying with `--no-public-ips`.

Vector pulls logs from Fly.io's internal log stream and doesn't need to accept inbound connections, so this:

- Removes the `[[services]]` / port 8686 block from the setup steps
- Deploys with `fly deploy --no-public-ips`
- Notes how to release IPs on already-exposed deployments (`fly ips list` / `fly ips release`)
- Adds a health check section using a top-level `[checks]` block against Vector's `/health` on 8686, which works without a public service or IP (the shipper can stop shipping logs without exiting, see #1828)

All of the above was verified on a live deploy; see the testing comments below.

Supersedes #930 (same fix, but that PR targeted a since-moved file and had gone stale) and #1828 (same health check goal, but via a public `[http_service]`, which this PR removes).

### DIFF
--- a/monitoring/exporting-logs.html.md
+++ b/monitoring/exporting-logs.html.md
@@ -46,7 +46,7 @@ fly secrets set LOGTAIL_TOKEN=<token provided by logtail source>
 
 You can configure as many providers as you'd like by adding more secrets. The secrets needed are determined by [which provider(s)](https://github.com/superfly/fly-log-shipper#provider-configuration+external) you want to use.
 
-The Log Shipper pulls logs from Fly.io's internal log stream and doesn't need to accept inbound connections, so it shouldn't have any public services or IP addresses. Before deploying, edit the generated `fly.toml` file and delete the entire `[[http_service]]` (or `[[services]]`) section.
+The Log Shipper pulls logs from Fly.io's internal log stream and doesn't need to accept inbound connections, so it shouldn't have any public services or IP addresses. Before deploying, edit the generated `fly.toml` file and delete the entire `[http_service]` (or `[[services]]`) section.
 
 Then deploy without allocating a public IP address:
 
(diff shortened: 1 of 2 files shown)


kcmartin:
**Tested this end to end against a real deployment** (throwaway app in my personal org, destroyed after):

- Followed the updated instructions verbatim: launched the shipper image, deleted the generated `[http_service]` section, deployed with `fly deploy --no-public-ips`. Result: no public IPs allocated, no DNS record created, port 8686 unreachable. flyctl confirms: "Your app is deployed but does not have a public or private IP address."
- Logs confirmed Vector connects outbound to the internal NATS stream, so the shipper genuinely needs no inbound services.
- Negative control: redeployed with the old-style public service config and allocated IPs. Vector's GraphQL playground was served publicly at `https://<app>.fly.dev/playground`, reproducing the exposure from the [original community report](https://community.fly.io/t/fly-log-shipper-publishes-graphql-publicly/14836).
- Remediation path: `fly ips release` on both addresses cleared the DNS records immediately; the hostname stopped resolving entirely within a few minutes.

Two things surfaced during testing and were folded into the PR: `fly launch` generates `[http_service]` (single brackets), so the text now matches, and `Datadog` was added to the Vale vocabulary to clear a pre-existing lint failure on this file.

kcmartin:
DONE: #1828 raises a related need (the shipper can wedge silently, and a health check was recommended by support). Since this PR removes the public service, the right approach is a top-level `[checks]` machine check against Vector's `/health` on 8686, which needs no service or public IPs. Plan is to add a short health check section here before this merges.

kcmartin:
Health check section added, using a top-level `[checks]` block against Vector's `/health` on 8686, so no public service or IP is needed. Verified on a live deploy: the check passes on a healthy shipper, goes critical when Vector hangs (simulated with SIGSTOP), and recovers. This covers the concern from #1828. Ready for review.

injoongy (approved):

PR #2426, opened 2026-07-09
remove kubernetes documentation

senyosimpson:
### Summary of changes

Remove Kubernetes documentation. We're sunsetting FKS as a whole, so we should remove the docs. Additionally, since basically no one uses it besides us, we don't have to go through a long deprecation cycle, hence just removing the docs wholesale.


machines/api/apps-resource.html.markerb
```diff
@@ -73,7 +73,7 @@ Machines must be associated with a Fly App. App names must be unique.
         | Property   | Type    | Required | Description                |
         |-------------|---------|----------|----------------------------|
         | `app_name`  | string  | yes      | Name of the app to create  |
-        | `enable_subdomains`  | boolean  | no      | Used for Fly Kubernetes. |
+        | `enable_subdomains`  | boolean  | no      | |
```
miguel-fly, line 76: Similar to other cases, we could add something like this

```suggestion
        | `enable_subdomains`  | boolean  | no      | If true, provisions a wildcard TLS certificate for `*.svc.<app-name>.fly.dev`, so services in the app can be served on per-service subdomains. |
```

miguel-fly (commented):
Nice cleanup! A couple of things I spotted:

**Leftover references**

- about/support.html.md#L102: FKS is still listed under "Supported Products." Should come out with the rest.
- flyctl/cmd/fly_extensions.md#L10: still links to the kubernetes extension docs this PR deletes, so that's a dead link now. Heads up though, this file (and the ones you already deleted under `flyctl/cmd/`) is auto-generated by the flyctl-bot sync. If the `kubernetes` subcommand is still in flyctl itself, the bot will just bring these files back next sync. Worth pulling the command out on the flyctl side too, or this fix won't stick.
- One more to consider: hiring/roles.html.md#L20 mentions "building a better, simpler K8s"

**Redirects**

Since we're deleting `/docs/kubernetes/*` entirely, anyone hitting those URLs from an old bookmark, search result, or the [FKS launch blog post](https://fly.io/blog/fks/) is going to get a 404. Might be worth adding a redirect, even just `/docs/kubernetes/* → /docs/`, rather than letting those links rot.

kcmartin:
Thanks, and sorry this sat. We're moving the docs to Mintlify on Sept 23rd, and everything here becomes .mdx, so this can't merge as written. I'm carrying it over: the content deletion, the leftovers miguel-fly found on about/support and fly_extensions, and redirects for the nine URLs, which this needs and doesn't have yet.

Two things I'd want from you before I do it: is the FKS sunset public yet, and is anyone removing the kubernetes commands from flyctl? The five generated pages here come straight back on the next release while those commands exist.

PR #2430, opened 2026-07-14
Retroactive review followup: Fix API details and style nits in warm pools guide

kcmartin:
Follow-up review fixes for the warm pools blueprint (#2427).

Technical:

- Use port 80 with the `http` handler for the Flycast service. Flycast is HTTP-only, and `.flycast` hostnames can't get public certs, so a `tls` handler on 443 would force clients to skip cert verification. This matches the config in the Connecting to User Machines blueprint.
- Allocate the Flycast address into the app's own network. The `network` field on the allocation doesn't default to the app's network; left empty, the address lands in the org's default network and can't reach a Machine on the app's isolated per-app network.
- Drop `min_machines_running` from the Machine config example; it only has an effect when autostop is enabled, and this config sets `autostop: "off"` deliberately.
- Name the specific `422` status for app-name collisions (verified against the live API by @kylemclaren), and add `422` to the Response Codes table in the Machines API reference so the guide cites a documented code.
- Note that the wait endpoint's 60s timeout is both the default and the max.

Links:

- Repoint the lease link to `#create-a-machine-lease` (the `#machine-leases` anchor doesn't exist), and point the exec link at the generated API reference on docs.machines.dev, since exec isn't in the hand-written reference yet.
- Use canonical `/docs/launch/processes/` instead of the `/docs/apps/processes/` redirect.

Style, matching other blueprints: sentence-case "Pointers and footguns" heading, colon separators in Related reading.

Added missing entry for this guide on the side navigation.

Not included here: bringing the hand-written Machines API reference pages up to date with `exec` and `ip_assignments` (both are documented in the generated reference at docs.machines.dev). That's left for a separate PR.

### DIFF
--- a/blueprints/warm-pool-user-machines.html.md
+++ b/blueprints/warm-pool-user-machines.html.md
@@ -43,7 +43,7 @@ curl -X POST "https://api.machines.dev/v1/apps/pool-x7f3k2/ip_assignments" \
 
 Pass the same `network` name you created the app with. The `network` field on the allocation targets which 6PN network the Flycast address routes within, and it does not default to the app's own network. Leave it empty and the address lands in your organization's default network, where it cannot reach a Machine that lives in the app's isolated per-app network.
 
-App creation fails if the name is already taken. Your worker will retry after crashes, so treat creation as idempotent: on a name-collision error, `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the failure was a real validation error, so rethrow it.
+App creation returns `422` if the name is taken. Your worker will retry after crashes, so treat that as idempotent: `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the `422` was a real validation error, so rethrow it.
 
 ## Track pool state in a database
 
--- a/blueprints/warm-pool-user-machines.html.md
+++ b/blueprints/warm-pool-user-machines.html.md
@@ -41,7 +41,7 @@ curl -X POST "https://api.machines.dev/v1/apps/pool-x7f3k2/ip_assignments" \
   -d '{"type": "private_v6", "network": ""}'
 ```
 
-App creation returns `422` if the name is taken. Your worker will retry after crashes, so treat that as idempotent: `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the `422` was a real validation error, so rethrow it.
+App creation fails if the name is already taken. Your worker will retry after crashes, so treat creation as idempotent: on a name-collision error, `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the failure was a real validation error, so rethrow it.
 
 ## Track pool state in a database
 
@@ -87,10 +87,9 @@ curl -X POST "https://api.machines.dev/v1/apps/pool-x7f3k2/machines" \
       "mounts": [{"volume": "vol_abc123", "path": "/data"}],
       "services": [{
         "internal_port": 8080,
-        "ports": [{"port": 443, "handlers": ["tls", "http"]}],
+        "ports": [{"port": 80, "handlers": ["http"]}],
         "autostart": true,
-        "autostop": "off",
-        "min_machines_running": 1
+        "autostop": "off"
       }],
       "checks": {
         "health": {"type": "tcp", "port": 8080, "interval": "15s",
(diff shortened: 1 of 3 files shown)


kcmartin:
Hey @kylemclaren, this is the follow-up from the retroactive review of your warm pools blueprint (#2427). The API usage is good overall (I checked it against the OpenAPI spec at docs.machines.dev). A few things that would be good if you can answer from your testing:

1. **App-name collision status code:** the guide said app creation returns `422` when the name is taken, but neither our response-code docs nor the OpenAPI spec mention 422. Did you observe that in your testing? If so we should add it to the reference; this PR softens the wording in the meantime.
2. **Wait endpoint timeout:** the guide called 60s the max, but everything we have documents 60s as the *default*. Did you hit a real server-side cap?
3. **`"network": ""` in the Flycast allocation call:** the app is created on a named network, so is the empty string intentional (meaning "use the app's own network")? If so, worth a line of prose saying that.

**Related**: `exec` and `ip_assignments` aren't in our Machines API reference at all, even though your guide leans on both and they're in the published API spec. Before we document them, can you (or the Machines folks) confirm they're considered public and stable?

kylemclaren (approved):
Amazing, thanks @kcmartin

blueprints/warm-pool-user-machines.html.md
```diff
@@ -101,7 +102,7 @@ curl -X POST "https://api.machines.dev/v1/apps/pool-x7f3k2/machines" \
     }
   }'
 
-# Wait for the VM to boot (max timeout is 60s)
+# Wait for the VM to boot (timeout defaults to 60s)
```
kylemclaren: I checked this against the live Machines API: 60s isn't only the default, it's the enforced maximum. `wait?...&timeout=61` returns `HTTP 400` with `invalid WaitMachineRequest.Timeout: value must be inside range [1s, 1m0s]` (and `timeout=60` succeeds). Worth noting, since a caller who wants to wait longer has to loop rather than raise the value.

```suggestion
# Wait for the VM to boot (timeout defaults to 60s, which is also the max)
```

blueprints/warm-pool-user-machines.html.md
```diff
@@ -34,14 +34,16 @@ curl -X POST "https://api.machines.dev/v1/apps" \
...
   -H "Authorization: Bearer ${FLY_API_TOKEN}" \
   -H "Content-Type: application/json" \
-  -d '{"type": "private_v6", "network": ""}'
+  -d '{"type": "private_v6", "network": "pool-x7f3k2"}'
 ```
 
-App creation returns `422` if the name is taken. Your worker will retry after crashes, so treat that as idempotent: `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the `422` was a real validation error, so rethrow it.
+Pass the same `network` name you created the app with. The `network` field on the allocation targets which 6PN network the Flycast address routes within, and it does not default to the app's own network. Leave it empty and the address lands in your organization's default network, where it cannot reach a Machine that lives in the app's isolated per-app network.
+
+App creation fails if the name is already taken. Your worker will retry after crashes, so treat creation as idempotent: on a name-collision error, `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the failure was a real validation error, so rethrow it.
```
kylemclaren: Optional: I reproduced the collision against the live API and it does return a specific code, `HTTP 422` with body `{"error":"Validation failed: Name has already been taken"}`. It's just absent from the response-code reference and OpenAPI spec (it's a Rails-style validation error). If we're comfortable citing an observed-but-undocumented code we can be specific here; if you'd rather stay conservative until it lands in the reference, your current wording is fine. Suggestion if we want to be specific:

```suggestion
App creation returns `422` if the name is taken. Your worker will retry after crashes, so treat that as idempotent: `GET /v1/apps/{name}` to confirm the app exists and reuse it. If it doesn't exist, the `422` was a real validation error, so rethrow it.
```

kylemclaren (commented):
Two follow-ups from verifying details against the live Machines API. Both are inline suggestions below.

kylemclaren:
@kcmartin comments 1 - 3 all addressed, thanks!

On exec and ip_assignments, those are documented:

- https://docs.machines.dev/machines/Machines_exec
- https://docs.machines.dev/apps/App_IPAssignments_create

blueprints/warm-pool-user-machines.html.md
kcmartin: Thanks for checking that!

blueprints/warm-pool-user-machines.html.md
kcmartin: Thanks for reproducing this! Since you've verified it against the live API, I'd rather be specific than vague, so I'm applying your suggestion. I've also added `422` to the Response Codes table in the Machines API reference.

PR #2447, opened 2026-08-03
Update docs for the Sentry partnership sunset

igor-ladkin:
New Sentry provisioning is now blocked platform-wide (superfly/web#4239) and removed from the dashboard/launch UIs (superfly/ui-ex#4992, #4993, #4996, #5005). This updates the docs to match:

- **`/docs/monitoring/sentry/`**: sunset warning banner up top — existing projects keep working, the page stays as reference for them, and new users are pointed at signing up with Sentry directly (same link as the API error message). Rest of the page untouched.
- **App handover guide**: Sentry handover now happens in the Sentry dashboard directly; new Fly-provisioned projects are gone.
- **Heroku migration**: New Relic row mirrors the SendGrid "keep using it" advice instead of suggesting `fly extensions sentry`.
- **Cell-based blueprint**: dropped the Sentry recommendation sentence.
- **Org roles**: trimmed Sentry from the extension-partner examples.

Deliberately untouched: the sidebar "Sentry Error Tracking" nav link (the page still serves existing customers) and the python/rust speedrun `fly launch` sample outputs (flyctl's own output, will self-correct when flyctl drops the field).

## Screenshots

deprecation-banner


igor-ladkin:
@codex review

stahnma:
I have a small preferences for dropping the "same result, one extra step" portion, but outside of that looks great.

stahnma (approved):

PR #2452, opened 2026-08-07
Finish the Sentry sunset cleanup

kcmartin:
Follow-up to #2447. That PR updated the Sentry page and four others, but three pages still described the partnership in present tense and sent new users toward a signup flow that no longer exists.

- **Going to production checklist:** the error tracking item told readers that "Fly.io organizations get a year's worth of Sentry Team Plan credits" and linked to the plan details anchor. This is the most likely path for a new user to hit the dead flow, since the checklist is what we point people at before launch. It now points at signing up with Sentry directly and setting `SENTRY_DSN` as an app secret.
- **Monitoring index:** dropped the "can claim a year's worth of Team Plan credits" sentence and the "from our extension providers" framing in the section heading. The linked Sentry page already carries the sunset banner.
- **Shared-nothing blueprint:** the closing recommendation linked to the community post announcing the integrated Sentry error tracking. Points to sentry.io instead. The other Sentry mentions in that blueprint are the author describing his own tooling, so they stay.

Two things outside this repo that are still open:

- `/docs/flyctl/extensions-sentry-create/` still reads "Provision a Sentry project for a Fly.io app" with no warning, and `fly extensions` still lists the subcommand. Those pages are generated from flyctl, so closing that path needs a change upstream.
- The `fly launch` output samples in the Next.js, Python, and Rust guides still show a `Sentry: false (not requested)` line. Accurate to what flyctl prints today, so left alone.

### DIFF
--- a/apps/going-to-production.html.markerb
+++ b/apps/going-to-production.html.markerb
@@ -97,7 +97,7 @@ Moving an app from staging to production can expose unexpected failure modes: se
   items: [
     
     { id: "metrics", title: "Monitor your app with fully-managed metrics", description: "Use managed Prometheus and managed Grafana dashboards to monitor your app. Read about [Metrics on Fly.io](/docs/monitoring/metrics/)."},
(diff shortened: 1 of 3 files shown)


igor-ladkin (approved):
LGTM

PR #2458, opened 2026-08-08
Fix launch command help typo

dane:
### Summary of changes

There is a typo in the CLI and the docs. I'm fixing both. It's just `CND` where I believe it should be `CMD`.


NullHypothesis:
Much appreciated!

PR #2467, opened 2026-08-21
Add new `--limit` flag.

NullHypothesis:
...that will soon be added to flyctl.


NullHypothesis:
Closing because interactive pagination in `flyctl` isn't that useful.

PR #2468, opened 2026-08-22
Change dashboard host from 0.0.0.0 to 127.0.0.1

sengeezer:
Updated the host address for the hermes dashboard command. It will refuse to bind to 0.0.0.0 for cases involving the method described here.

### Summary of changes

Updated bind address as per hermes error message.

### Notes

Error message:

```
→ Using web dist from HERMES_WEB_DIST: /opt/hermes/hermes_cli/web_dist
Refusing to bind dashboard to 0.0.0.0 — the auth gate engages on non-loopback binds, but no auth providers are registered.

Configure an auth provider before exposing the dashboard:
  • Password: set dashboard.basic_auth.username + password_hash in config.yaml
    (hash with: python -c "from plugins.dashboard_auth.basic import hash_password; print(hash_password('your-password'))")
  • OAuth: run `hermes dashboard register` (Nous Portal) or install a DashboardAuthProvider plugin.
There is no unauthenticated public-bind option — to keep it local, bind 127.0.0.1 and tunnel in (SSH / Tailscale).
Error: ssh shell: Process exited with status 1
```


kcmartin:
Thanks for catching this. The dashboard instructions were rewritten in #2482 on 15 September, which fixes the same bind failure a different way: `[::]` with `HERMES_DASHBOARD_WS_HOST=127.0.0.1`, plus a note about dashboard auth.

Closing this as superseded, but please say so if you still hit the problem with the new instructions and I'll get it fixed. One heads up: our docs move to a new platform next week, so the file will be .mdx after that.

PR #2473, opened 2026-08-31
Add backup CT search link

rellem:
crt.sh is a great resource, but it seems to have server issues quite often so a backup link could be useful.

Disclaimer: I'm the founder of CertObserver.


kcmartin:
Thanks for this, and for the upfront disclosure, we appreciate the transparency.

We're going to pass for now: we keep this list to tools the team reaches for ourselves, and crt.sh covers the same job. If CertObserver becomes one we use, I'd be glad to add it then.

PR #2484, opened 2026-09-15
docs(mpg): correct Ecto connection lifetime guidance

miguel-fly:
### Summary of changes

`max_lifetime` shipped in DBConnection 2.10.0 (2026-04-24), with an idle-connection fix in 2.10.1. The client-configuration page still told Elixir users it did not exist and called it a known gap, leaving Ecto as the only client we gave no connection-lifetime advice for. Adds it to both the client-configuration and Phoenix guide examples.

The same note claimed PgBouncer's `server_lifetime` defaults to 3600s. MPG sets 600 in its own template, so the note overstated the fallback protection by 6x. The replacement names the two settings that actually apply: `client_idle_timeout` (600s, closes idle client connections) and `server_lifetime` (600s, retires pooler-to-Postgres connections).

Separately, the Phoenix guide showed `timeout: 15_000` with no explanation. That is Ecto's query timeout and has been read as an idle timeout by both customers and support.

Four places justified the 600s recommendation as "recycle before proxy timeout" or "before the proxy needs to kill them". The proxy's drain behaviour is not something these docs should be asserting, and the figure they implied is out of date. The recommendation is unchanged; the reasoning now points at our own PgBouncer settings, which are in the cluster template and verifiable.

Reported by a customer on T-19471.

### Preview
Screenshot 2026-09-15 at 4 02 10 PM

Screenshot 2026-09-15 at 4 02 49 PM

Screenshot 2026-09-15 at 4 03 11 PM

Screenshot 2026-09-15 at 4 03 24 PM

Screenshot 2026-09-15 at 4 03 48 PM


Roadmaster (approved):
LGTM thanks!

